PR #12, opened 2015-06-16
Filter decision

kencall:
I have added a EDProducer that can extract HLT and MET Filter decision bits from the miniAOD file and add them as booleans to the tree produced by TreeMaker.

This includes a new module in Utils, and an updated makefromMiniAOD that understands how to use the producer.

Ken Call


kencall:
Further updates have mooted this PR.

PR #13, opened 2015-06-16
Filter decision

kencall:
I have added an EDProducer that can extract HLT and MET Filter decision bits from the miniAOD file and add them as booleans to the tree produced by TreeMaker.

This includes a new module in Utils, and an updated makefromMiniAOD that understands how to use the producer.

Ken Call


awhitbeck:
Hi Ken, 

This looks good. I would like to suggest that for this and for other developments, we should start to separate things into separate configuration files so that things are easier to read.  Is it possible to move the setup you have in the main config, makeTreeFromMiniAOD_cff.py, to a cff in the python directory and then just load this.  Also I think its possible to just pass the python list to the config file without doing this thing where you make a python list and the extended each element into a cms.vstring().   This is a minor point, but simpler is better.  

Thanks!
Andrew

kencall:
Andrew,

I agree about simpler being better. I'll clean it up a bit. I agree with you comments concerning the config file. Perhaps we could clean up makeTreeFromMiniAOD_cff.py up a bit to the point where it simply loads the appropriate cff's for whatever branches we are interested in.

Ken C.

On 06/19/2015 04:30 AM, Andrew Whitbeck wrote:

Hi Ken,

This looks good. I would like to suggest that for this and for other developments, we should start to separate things into separate configuration files so that things are easier to read. Is it possible to move the setup you have in the main config, makeTreeFromMiniAOD_cff.py, to a cff in the python directory and then just load this. Also I think its possible to just pass the python list to the config file without doing this thing where you make a python list and the extended each element into a cms.vstring(). This is a minor point, but simpler is better.

Thanks!
Andrew

awhitbeck:
Hi Ken, 

Sounds great.  I agree, we should organize the main config file a bit.  Frank had the same idea, you and Frank should coordinate on this type of work. 

Thanks!
Andrew

PR #17, opened 2015-07-09
Cmssw 7 2 x

ahmadborzou:
Hi, 
This is to include the Hadronic tau oriented code. Basically this part of the code can be switched off if wanted. So one can leave the switch off and nothing will change for him/her.


hatakeyamak:
Hi TreeMaker admins,

I just wanted to follow up on this PR with a little more information on the changes.

[1] AllHadronicSUSY/TreeMaker/interface/HadTau_TreeMaker.h
[2] AllHadronicSUSY/TreeMaker/src/HadTau_TreeMaker.cc
[3] AllHadronicSUSY/TreeMaker/python/hadtau_treeMaker.py
We would like to use these instead of standard TreeMaker.cc & TreeMaker.h in order to
store some information in vectors into trees. Another possibility may be that
regular treemaker.cc/h/py pick up these changes, but at this point our preference is
to add these files/module.

[4] AllHadronicSUSY/Utils/python/alljetsproducer_cfi.py
[5] AllHadronicSUSY/Utils/src/AllJetsProducer.cc
These are used to store reclustered jets that can be extended below 10 GeV. The code is written
in such a way that above the 10 GeV threshold, it reads slimmedJets from MiniAOD, and it uses
reclustered jets below 10 GeV. This will add a new list of jets “slimJetXXX”.

[6] AllHadronicSUSY/TreeMaker/python/makeTreeFromMiniAOD_cff.py
[7] AllHadronicSUSY/TreeMaker/test/runMakeTreeFromMiniAOD_cfg.py
These configs have associated changes. A new switch TauHad is introduced in [6] which
is false by default.

[8] AllHadronicSUSY/Utils/src/GenLeptonRecoCand.cc
This is modified in order to store the tau neutrino information which we need to
figure out the tau visible decay four-vector.

I understand that a lot of work is on-going, but I will appreciate if you can consider this as important additions to make.

Thanks.

awhitbeck:
Hi Ken, Ahmad,

I don't understand.  We are already storing vectors in the trees with the original tree maker.  Currently, arrays of pt, eta, phi, and E can be stored or vectors of TLorentzVectors can be stored.  This is already duplicated the same basic functionality and it would be much better if we didn't add a third way to save information to the trees.  I would really like to encourage you guys to conform to one of the two way to save four vectors that are already implemented.  

Also, doesn't this mean that the hadronic tau information will be stored in a totally separate tree?  If so, I don't understand the point.  

For [5] why are we just doing this with a standard python module from the jet/MET group.  It seems like you are basically rewriting modules that exist and are maintained by experts.   Am I missing something here?

Overall, I have some concerns about how messy the code is getting.  I would like to keep things as organized as possible so that things remain manageable as more and more analysis tools are added. 

Andrew

kpedro88:
Hi Ken, Ahmad,

I agree with Andrew's concerns and was just about to write a similar comment. This PR is not in good shape to be merged right now.

Ahmad, if you were having trouble storing TLorentzVectors in the tree, your problems might have been fixed by the updates I have made today. I encourage you to save your current work in a separate branch, update your version of this branch (CMSSW_7_2_X), and see if that works better.

-Kevin

hatakeyamak:
Hi Kevin, Andrew,

Thank you for your inputs. And, thanks Kevin for help/suggestions via private communications. I think we will be making an update or new PR without [1-3].

I think this producer [5] is necessary. Reclustering is done based on the standard procedure. My description was not quite accurate, but as Hongxuan mentioned in email, in order to avoid making this list of jets too big, the reclustered low pt jets are stored only if they are matched to gen leptons, because those are low pt jets we are interested in, and this is what this producer is doing.

Ken

PR #26, opened 2015-07-14
FilterFlags and TriggerResults is smarter

kencall:
FilterFlags can now be set to locate the product using default values
TriggerResults can now match substrings. It will search for TriggerNames that contain the passed names as substrings.

The current behaviour is preserved if the entire trigger name is passed; however, it is now possible to pass a trigger name fragment, it will search for any triggers names that contain the fragment as a substring.

### DIFF
--- a/AllHadronicSUSY/Utils/python/filterdecisionproducer_cfi.py
+++ b/AllHadronicSUSY/Utils/python/filterdecisionproducer_cfi.py
@@ -3,6 +3,6 @@ import FWCore.ParameterSet.Config as cms
 filterDecisionProducer = cms.EDProducer('FilterDecisionProducer',
 trigTagArg1  = cms.string('TriggerResults'),
 trigTagArg2  = cms.string(''),
-trigTagArg3  = cms.string('PAT'),
+trigTagArg3  = cms.string(''),
 filterName    =   cms.string("Flag_METFilters")
 )
--- a/AllHadronicSUSY/Utils/src/FilterDecisionProducer.cc
+++ b/AllHadronicSUSY/Utils/src/FilterDecisionProducer.cc
@@ -86,8 +86,19 @@ FilterDecisionProducer::FilterDecisionProducer(const edm::ParameterSet& iConfig)
   trigTagArg1_ = iConfig.getParameter <std::string> ("trigTagArg1");
   trigTagArg2_ = iConfig.getParameter <std::string> ("trigTagArg2");
   trigTagArg3_ = iConfig.getParameter <std::string> ("trigTagArg3");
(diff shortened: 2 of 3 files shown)


kpedro88:
Thanks for adding these features. A few questions/comments:
1. We had put a patch in makeTreeFromMiniAOD_cff to fix the filter tags (the "filtertag" option, which could be set to "PAT" or "RECO" as needed). Is this no longer necessary for the filters? If so, it should be taken out of those configurations (although the flag itself is also used elsewhere, so it will need to be retained for now).
2. The trigger names in the config list still have the _vX part: https://github.com/TreeMaker/TreeMaker/blob/CMSSW_7_2_X/AllHadronicSUSY/TreeMaker/python/makeTreeFromMiniAOD_cff.py#L378-396. If I understand your change correctly, these should be removed so that any version can be matched.

Let me know if my understanding is correct on both of these points. I can make these changes when merging the PR, or if you want to make them, just push your commit to this branch and the PR will update automatically.

kencall:
Hi Kevin,

Your understanding on both points is correct. I left the filtertag option alone as it could useful.

With regards to the trigger names, I tested it with the version numbers removed; however, I didn't want to make changes to the expected default behavior without input from others. With the entire trigger name specified, the default behavior is the same as before.

Ken Call

On 07/14/2015 12:19 PM, Kevin Pedro wrote:

Thanks for adding these features. A few questions/comments:
1. We had put a patch in makeTreeFromMiniAOD_cff to fix the filter tags (the "filtertag" option, which could be set to "PAT" or "RECO" as needed). Is this no longer necessary for the filters? If so, it should be taken out of those configurations (although the flag itself is also used elsewhere, so it will need to be retained for now).
2. The trigger names in the config list still have the _vX part: https://github.com/TreeMaker/TreeMaker/blob/CMSSW_7_2_X/AllHadronicSUSY/TreeMaker/python/makeTreeFromMiniAOD_cff.py#L378-396. If I understand your change correctly, these should be removed so that any version can be matched.

Let me know if my understanding is correct on both of these points. I can make these changes when merging the PR, or if you want to make them, just push your commit to this branch and the PR will update automatically.

kpedro88:
If we remove the version numbers in the trigger names, the only change will be that triggers with different version numbers can also be included, right? As long as all the triggers people expect will be included, I don't think it's a problem to potentially include some other ones.

kencall:
That is correct.

Ken

On 07/14/2015 04:49 PM, Kevin Pedro wrote:

If we remove the version numbers in the trigger names, the only change will be that triggers with different version numbers can also be included, right? As long as all the triggers people expect will be included, I don't think it's a problem to potentially include some other ones.

PR #31, opened 2015-07-17
Additions/changes for hadtau studies

hatakeyamak:
This is an revised pull request of what Ahmad submitted last week.

[1] AllHadronicSUSY/Utils/python/jetsforhadtauproducer_cfi.py
[2] AllHadronicSUSY/Utils/src/JetsForHadTauProducer.cc

These are used to store reclustered jets that can be extended below 10 GeV. The code is written
in such a way that above the 10 GeV threshold, it reads slimmedJets from MiniAOD, and it uses
reclustered jets below 10 GeV. For jets below 10 GeV, we store them only if they match gen or reco leptons. This will add a new list of jets “slimJetXXX”. We renamed the files to more clearly indicate the purpose of the code.

[3] AllHadronicSUSY/TreeMaker/python/makeTreeFromMiniAOD_cff.py
[4] AllHadronicSUSY/TreeMaker/test/runMakeTreeFromMiniAOD_cfg.py
These configs have associated changes. A new switch hadtau is introduced in [3-4] which
is false by default.

[8] AllHadronicSUSY/Utils/src/GenLeptonRecoCand.cc
This is modified in order to store the tau neutrino information which we need to
figure out the tau visible decay four-vector. Also, we improved the definitions of tau decay daughters to store (we store down to either stable particles or pi0) by consulting with Arne.


kpedro88:
This looks good at a glance. I'll test it in the morning.

One quick thing: could you add an output message like [this one](https://github.com/hatakeyamak/TreeMaker/blob/CMSSW_7_2_X/AllHadronicSUSY/TreeMaker/test/runMakeTreeFromMiniAOD_cfg.py#L25) for the hadtau parameter in the test cfg?
It's useful for me to keep track of the parameters when I run tests.

hatakeyamak:
Thanks. I just added this flag you have in mind. I checked these additions under both 72x (Phys14 MC) and 74x (data and Spring15 MC), so hopefully merging will go smoothly.

kpedro88:
When I merge this commit locally, I get a warning from Python:

```
src/AllHadronicSUSY/TreeMaker/python/makeTreeFromMiniAOD_cff.py:6: SyntaxWarning: import * only allowed at module level
```

Could you change [line 411](https://github.com/hatakeyamak/TreeMaker/blob/CMSSW_7_2_X/AllHadronicSUSY/TreeMaker/python/makeTreeFromMiniAOD_cff.py#L411) to import only what you need?

hatakeyamak:
Sure. I think this is done now.

PR #33, opened 2015-07-17
fix bug in storage of GenTauNu candidates

kpedro88:
@hatakeyamak, I just noticed this bug in how the GenTauNu candidates are stored. This PR will fix it.


hatakeyamak:
Oops. Thank you. I didn't appreciate the differences between "," and "|". Probably now I understand that "|" works when the length is expected to be the same.

kpedro88:
Yes, essentially; "|" is intended to store an additional variable for a given candidate. For example, in the line modified in this PR, the TauDecayCand_pdgID is stored alongside the TauDecayCand RecoCand. "I_pdgID" is used to indicate that the variable type is an integer. GenTauNu is a set of candidates separate from TauDecayCand, so it needs to be added as a separate item in the RecoCandVector.

hatakeyamak:
Okay. Thanks. Now, I know it!

PR #35, opened 2015-07-24
TreeMaker code upgrades

kpedro88:
This pull request (and the corresponding new branch Run2 in the main repository) includes several major upgrades to the TreeMaker code:
1. The actual TreeMaker producer has been entirely rewritten to use class templates for organizing the branches to be added to the output tree. In the process, some features were changed:  
   a) variable-length arrays no longer used for any branches  
   b) RecoCands stored by default as vector&lt;TLorentzVector&gt;  
   c) Option is available to store RecoCands as four vector&lt;double&gt; of pt, eta, phi, energy (instead of vector&lt;TLorentzVector&gt;)  
   d) "additional variables" stored alongside RecoCands no longer used - the existing vector&lt;bool,int,double&gt; functionality should be used for these branches  
   e) Branches are sorted alphabetically (case-insensitive) by default, so that branches of different types with similar names (i.e. from the same collection/producer) are kept together in the tree structure  
   f) Option to switch back to standard branch sorting (based on branch type)
2. The main TreeMaker configuration, makeTreeFromMiniAOD_cff.py, has been reorganized to separate the optional background estimation customizations into their own .py files which are imported if desired. The indentation and comments in the main cff has also been standardized to improve readability. Let's try to maintain these standards going forward. The different background estimation .py files are: doHadTauBkg, doTagAndProbe, doLostLeptonBkg, doZinvBkg. There is also doProcessesForLLTP which contains processes that are used by both lost lepton and tag and probe.
3. The directory structure of the repository was reorganized to eliminate the unnecessary "AllHadronicSUSY" directory and simplify the `git clone` procedure. (This has the unfortunate side effect of disconnecting the file histories on GitHub, but all the commits can still be browsed to see the incremental changes.)
4. The WeightProducer python configuration was rewritten to use a simple data structure to store the info for each sample (name, cross section, number of events, etc.). This reduces code duplication and streamlines the configuration.
5. A readme file was added, which contains the installation recipe and instructions, and an overview of the configuration parameters. The goal is to track the installation recipe in Git along with the code, so that any changes are obvious to users (rather than using Twiki, which is not synced with the code version).
6. Some parameter names were synchronized between different configurations (minor change).

I have run basic tests on these upgrades and it appears that everything works as expected. However, analyzers should test the new version for themselves and make sure that nothing was broken or lost (especially for the different background estimations).

I will leave this pull request open for several days to allow comments from the other developers.

To test this pull request, you can clone from my fork:

```
git clone git@github.com:kpedro88/TreeMaker -b upgrade
```


kpedro88:
I added a bunch of commits today covering many of the discussion items and future work from yesterday's meeting, including:
- Store JetsProperties for all GoodJets by default
- Remove unneeded T&P ROOT files
- Default lumi for WeightProducer = 1 pb-1
  - as an aside, WeightProducer.cc could use a good cleaning at some point
- WeightProducer now accounts for negative weight events in "constant" mode
- Add Spring15 samples to WeightProducer cff (from commissioning twiki)
- Update b-tagging CSV working point for 74X
- Add prescales to trigger producer

@kencall @sbein take a look at the last commit and make sure it looks okay to you. I streamlined the input tag creation a little bit in the process of adding the prescales. I also changed the branch name of PassTrigger to TriggerPass, just to organize the tree branches so all the Trigger ones get sorted together, and the type to a boolean (as I previously discussed with Ken).

sbein:
Hi Kevin,
Thanks for all the heroic re-organizing, that will be appreciated in the
long run! I didn't catch the conversation about changing PassTrigger to a
boolean. I originally wrote it as an int that takes the value -1 in the
case a trigger path failed to be recorded. I guess you decided this was an
overly conservative contingency?
Sam

kpedro88:
Hi Sam,

The MET filters still use that form, but for the trigger decisions, it's no longer necessary. A while ago, we changed the producer to output the decisions in a single vector branch. Now, if a trigger path fails to be recorded, it just won't be added to the vector.

-Kevin

kpedro88:
Hi Sam,

The MET filters still use that form, but for the trigger decisions, it's no
longer necessary. A while ago, we changed the producer to output the
decisions in a single vector branch. Now, if a trigger path fails to be
recorded, it just won't be added to the vector.

-Kevin

sbein:
Hi Kevin,
I noticed that producer got intelligently modified a couple of times.
Everything looks good to me. I'll adapt to any unexpected changes.
Thanks,
Sam

PR #41, opened 2015-08-17
update on eras, primary vertex definition, HBHE noise filter config, addition of a few HLT

hatakeyamak:


kpedro88:
Hi Ken,

It looks like you figured out some fixes for the HBHE noise filter, which is good. Just before you sent this, I was looking at the configuration and noticed that we weren't setting the era properly.

Just one small typo that I noticed:
https://github.com/hatakeyamak/TreeMaker/blob/CMSSW_7_2_X/AllHadronicSUSY/Utils/src/PrimaryVerticesInt.cc#L135
It should be `if(fabs(vitr->z()) >= maxZ_)`, to be consistent.

Just a note for myself (or any other enterprising developers): Now that PrimaryVerticesInt is doing the quality check, we should update TrackIsolationFilter to use the result of this, instead of redoing the quality check (several times). This probably doesn't matter for the commissioning branch, so I'll add it to my list of updates for the Run2 branch.

-Kevin

hatakeyamak:
Hi Kevin,

Thanks for checking.
https://github.com/hatakeyamak/TreeMaker/blob/CMSSW_7_2_X/AllHadronicSUSY/Utils/src/PrimaryVerticesInt.cc#L136
Maybe this line should be
            if(fabs(vitr->position().Rho())>maxRho_)  continue;
in order to require
!isFake && ndof > 4 && abs(z) <= 24 && position.Rho <= 2"

If my implementation can be improved, please feel free to overwrite my changes. I just did this PR so that I won't forget.

Cheers,
Ken

awhitbeck:
Hi ken, 

thanks for organizing the updates. 

About the good vertex selection.  I would actually prefer that we use the standard vertex selector and the cutstringparser so that its easier to see what is going on from the config file.  This is actually a more general comment I have, but since we're updating this part of the code:

```
process.goodVertices = cms.EDFilter("VertexSelector",
                                    src = cms.InputTag("offlineSlimmedPrimaryVertices"),
                                    cut = cms.string("!isFake && ndof > 4 && abs(z) <= 24 && position.Rho <=2"),
                                    filter = cms.bool(False)
                                    )
```

Then we should pass goodVertices to usual producers.

hatakeyamak:
I think it's good to use VertexSelector, but I personally don't know how to use this to store Nvtx (not just a bool for events with Nvtx>=1) which we wanted to use for pileup reweighting this time. If this can be done, I think it's certainly good to use a more centralized tool. I just made a PR so that these updates won't be forgotten, but please feel free to overwrite mine with a better solution :)

kpedro88:
Hi Ken,

I'm going to close this and send another PR. I cherry-picked your commit to get the changes as you made them (except for the vertex stuff, which I rewrote to use the selector).

-Kevin

hatakeyamak:
Sounds good. Thanks.

Ken

PR #44, opened 2015-09-01
Lost lep run2

jbradmil:
Added gen lepton TLVs to lost lepton trees, removed unnecessary PF candidates.


kpedro88:
A few comments:
1. In doLostLeptonBkg.py, when you're removing processes and collections, I would prefer if you just deleted them instead of commenting out the lines. It can get confusing in the future when we don't remember why something is commented out. It's easy enough to revert the change with the tracked history in Git if we decide we want to put things back.
2. Related to the above, it looks like most of the processes in doProcessesForLLTP.py are only used in the tag & probe part now... but we can leave it for the time being.
3. In general, it is quite frustrating to look through a diff when 95% of it is whitespace formatting changes. There are options to suppress these changes so they don't appear in the diff, but nevertheless, try to avoid this in the future.

PR #46, opened 2015-09-06
QCD minDeltaPhi

aminghiasi:
Hi Kevin,

Pawandeep has talked with you about including our module into the official TreeMaker repo. This is the first time i send a pull request. Please let me know if there is anything wrong with my request.

Best,
Amin


kpedro88:
Hi Amin,

A few comments:
1. This PR adds a large number of branches to the tree. I would prefer to store things more concisely, using vector&lt;TLorentzVector&gt; for pt/eta/phi/E, and vector&lt;double&gt;, vector&lt;int&gt;, etc. for storing related quanties (PDG ID, CSV, etc.).
2. It also seems like a lot of the jet information is duplicated (since we already store all the GoodJets: https://github.com/TreeMaker/TreeMaker/blob/Run2/TreeMaker/python/makeTreeFromMiniAOD_cff.py#L614). Would it be possible just to use the existing stored GoodJets info, or are you applying special selection requirements?
3. The DeltaPhiQCD process should use the METTag variable from the main cff, because the MET collection name can change if the corrections are updated. Example: https://github.com/TreeMaker/TreeMaker/blob/Run2/TreeMaker/python/makeTreeFromMiniAOD_cff.py#L672
4. It would be helpful for future maintenance if the indentation in DeltaPhiQCD.cc could be standardized... right now it is a mix of tabs and spaces, which makes it quite difficult to follow. (This is lower-priority than the other items.)

-Kevin

aminghiasi:
Hi Kevin,

1- I merged a lot of variables in vectors. Now the number of branches I'm adding is much smaller. We didn't merge minDeltaPhi values as we use them a lot. So we want it to be easy to compare different minDeltaPhi values. 
2- Actually, we have a different selection requirement. We apply |eta| < 5 but I think the current requirement is |eta|<3
3- Done!
4- I tried to have a more organized code

Regards,
Amin

kpedro88:
Hi Amin,
1. This is good progress. Are you sure that the minDeltaPhi values could not be put into vectors? It is still quite easy to compare them in the tree. For example, if you had a vector&lt;double&gt; RJetMinDeltaPhiEta5, with index 0 = Star, index 1 = Njle3, index 2 = Njle4, index 3 = Njle5, then you could compare them in TTree::Draw() with RJetMinDeltaPhiEta5[0], RJetMinDeltaPhiEta5[1], etc. This would cut down the number of MinDeltaPhi branches by a factor of 4, so it would be quite appreciated.
2. The GoodJets already include all jets with |eta| < 5, so if you are not applying any additional requirements, it would be better just to use the jets already stored in the tree. For the GenJets, you can still store them separately for now. (Eventually, I might add central storage for GenJets and their properties.)
3. Great! I just noticed another parameter that needs to be passed through: in the 74X version, there is a different name and cut value for the b-tag discriminator: https://github.com/TreeMaker/TreeMaker/blob/Run2/TreeMaker/python/doZinvBkg.py#L68-L73. However, if you can switch to using the already-stored GoodJets collection, you will not need to deal with this.
4. If you look at https://github.com/TreeMaker/TreeMaker/pull/46/files, the spacing is still rather inconsistent... I also noticed some places where the code could be made more efficient, but I can address this in a subsequent pull request.

-Kevin

aminghiasi:
Hi Kevin,
1. I put minDeltaPhi variables in vectors. ( 4 vectors, 1 for eta<2.4 and one for eta < 5 and two other vectors show what every index of every vector is)
2. I am not sure what variables are in common between official TreeMaker any our module. An important point is that we save eight jets while in the TreeMaker, usually the first 3 jets are saved. To summarize, I am not sure what variables are getting calculated twice. 
3. Done!
4. The code looks organized when I open it using vim. This is strange! I think if you download and open the code in Linux or mac, the spacing inconsistency will disappear. At least this is what happens to me!
- Amin

kpedro88:
Hi Amin,

1. I had thought of doing this in a slightly different way: vector&lt;double&gt; minDeltaPhiEta24 and vector&lt;int&gt; minDeltaPhiIndexEta24 separately (and similarly for Eta5). Generally, it is not preferable to store an int as a double. I had not considered the separate "names" vector, but if this is useful for you, that's fine. With the two vectors minDeltaPhiEta24 and minDeltaPhiIndexEta24, the "names" vector would only need to have half as many entries. The entries could probably be simplified to "Star", "Njle3", "Njle4", "Njle5"; then there would only need to be one vector&lt;string&gt; minDeltaPhiNames that would label all the vectors. Actually, now I notice you've combined the reco and gen variables into one vector. I think it's fine to have separate reco and gen vectors for clarity.
2. All of the GoodJets are saved in the TreeMaker, along with many properties: https://github.com/TreeMaker/TreeMaker/blob/Run2/TreeMaker/python/makeTreeFromMiniAOD_cff.py#L614-L628. If you can use these branches instead of duplicating them, that would be best.
3. Another typo I just noticed in DeltaPhiQCD.py: "DeltaPhiQCD:RJetLorentzVector" is added twice. I think the second one should be DeltaPhiQCD:GenJetLorentzVector.
4. This is because vim is set to display a tab character as 8 spaces, whereas I have Github set to display a tab character as 4 spaces. In general, code should use only spaces or only tabs for indentation, to avoid this potential mismatch.

-Kevin

aminghiasi:
Hi Kevin,

I changed my code according to your comments. Please let me know if I still need to change my code.

Best,
Amin

kpedro88:
Hi Amin,

This mostly looks good. I would still prefer not to store this:
process.TreeMaker2.VectorTLorentzVector.extend ( [ 'DeltaPhiQCD:RJetLorentzVector' ] )
because it duplicates information.

-Kevin

aminghiasi:
Hi Kevin,

I changed the code and removed RJetLorentzVector

Regards,
Amin

PR #47, opened 2015-09-11
minDeltaPhiQCD

aminghiasi:
Hi Kevin,

Apparently Owen want this module to be enabled in the central production campaigns. In other words, he doesn't want to run his own crab productions in the future and be able to use the central production. I think you already know about it. 
For this reason, I changed the default value of our QCD variable to True. Please let me know if there is a better way to realize what Owen wants.

Regards,
Amin


kpedro88:
Hi Amin,

As a matter of design, I have all the background estimation variables set to "False" by default in makeTreeFromMiniAOD_cff.py. When I merge my new production code (hopefully in the next few days), all the background estimations will be set to "True" by default in runMakeTreeFromMiniAOD_cfg.py, as we at FNAL already agreed to do a central production for all the background estimators:

https://github.com/kpedro88/TreeMaker/blob/upgrade/Production/test/runMakeTreeFromMiniAOD_cfg.py#L11-L15

-Kevin

aminghiasi:
Hi Kevin,

Thank you very much for your explanation

PR #51, opened 2015-09-15
better integration for DeltaPhiQCD, cleanup and fix a few bugs

kpedro88:
@aminghiasi, I noticed several places where the DeltaPhiQCD code could be better integrated with the rest of TreeMaker or written more efficiently. In the process, I fixed a few bugs.
1. The two categories of jets are: (pT > 30, |eta| < 5.0) and (pT > 30, |eta| < 2.4). We already produce these categories as MHTJets and HTJets, respectively, so it's easier just to use those collections instead of redoing all the cuts. I generalized SubJetSelection to make the same collections for GenJets, also (and then we can just add the GenMHTJets directly to the tree).
2. I added GenMET to the METDouble producer, and made GenMHT from the GenMHTJets collection, further simplifying the DeltaPhiQCD code. (I also fixed a bug where the GenMET didn't get filled when the MET recorrection tool was used.)
3. I simplified the calculation of MHTStar: just calculate the MHT vector from all the jets, then add back each jet separately.
4. The vectors of MinDeltaPhi results can be set using initializer lists (new C++11 feature).
5. It's easier to sort the gen neutrinos by pT using a map, and then just take the highest five entries (using reverse iterators, because maps sort from low to high).
6. Bug fix: for genetacut == 2.4, the Eta5 results vectors were filled again (typo), so the Eta24 vectors were empty.
7. The MHTStar was calculated using only the pT > 30 jets, but it was calculated _for_ all the jets. This seems like it probably wasn't what was intended. Now, using the subjet collection, it can only consider the jets with pT > 30 and the desired eta cut.
8. I wasn't sure if this was a bug, so I didn't address it yet: the mindeltaphiXjetindex is stored as "index+1", where "index" is the collection index. It seems like the +1 shouldn't be there, but maybe I misunderstand how you use the index.

I tested the new code with 100 QCD events and didn't see any unexpected changes. Let me know if I misunderstood anything and if I should fix point 8 before merging.

### DIFF
--- a/Production/test/runMakeTreeFromMiniAOD_cfg.py
+++ b/Production/test/runMakeTreeFromMiniAOD_cfg.py
@@ -32,7 +32,21 @@ jsonfile=parameters.value("jsonfile",scenario.jsonfile)
 jecfile=parameters.value("jecfile",scenario.jecfile)
 residual=parameters.value("residual",scenario.residual)
 era=parameters.value("era",scenario.era)
-    
+
+# The process needs to be defined AFTER reading sys.argv,
+# otherwise edmConfigHash fails
+import FWCore.ParameterSet.Config as cms
+from Configuration.StandardSequences.Eras import eras
+process = cms.Process("RA2EventSelection")
+if era=="Run2_25ns":
+    process = cms.Process("RA2EventSelection",eras.Run2_25ns)
+elif era=="Run2_50ns":
+    process = cms.Process("RA2EventSelection",eras.Run2_50ns)
+
+# configure geometry & conditions
+process.load("Configuration.StandardSequences.GeometryRecoDB_cff")
+process.load("Configuration.StandardSequences.MagneticField_AutoFromDBCurrent_cff")
+
 # Load input files
 readFiles = cms.untracked.vstring()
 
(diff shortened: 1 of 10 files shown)


aminghiasi:
Hi Kevin,

Thank you so much for spending time fixing my code. Basically you rewrote my code completely! I am reading what you did and learning from it.

Regarding number 8: I think it should be 'Index+1'. Index variable counts 0, 1, 2, 3, 4, 5, 6, 7 therefore when index == 0, it is the first jet, when index == 1, it is the second jet and so on. Another way is to change index variable in a way that it counts 1 to 8 instead of 0 to 7. Does what I'm saying make any sense? Please let me know if I'm making a mistake

Thank you very much,
Amin

kpedro88:
At first I was just going to make the MHTstar calculation faster, but then I noticed how some of the collections and quantities could be integrated better with what TreeMaker already does. I'm glad it's instructive!

I see what you're saying about the index. I guess it depends on how you want to use it. If you want to check the properties of the jet with that index, I think it's better to keep the 0-7 count, since the first jet in the ntuple vectors will have index 0. (Of course, since we store the properties for all GoodJets, you'd have to apply the pT and eta cuts again as well.)

aminghiasi:
Hi Kevin,

You are more experienced than me. You decide what choice is better.

kpedro88:
If the current version works for you, then let's leave it for now.

However, I noticed that the mindeltaphistarindex is set to just i, not i+1. Should I change this to be consistent with the others?

aminghiasi:
Hi Kevin,

Sure! You are right. It should be i+1 instead of i

Thank you very much

PR #66, opened 2015-10-02
NonpromptPhotons added

bmahakud:
Non prompt photon collections are added to the PhotonIDIsoProducer.cc


kpedro88:
@bmahakud - This mostly looks fine to me, just one thing needs to be changed:
The new code in PhotonIDisoProducer.cc (lines 268-310) uses the gen particles collection, which is not available in data. The whole section should be surrounded by this condition:

```
if (genParticles.isValid()){

}
```

Otherwise, it will crash when running on data. If you can update the pull request with this (just push the commit to your remote branch and GitHub will pick it up automaticaly), I'll merge it for the production.

kpedro88:
@bmahakud - Also, I assume the "hadronization" stuff is deliberately commented out for now?

bmahakud:
Hi Kevin,
Yes that is commented out for now..I will fix it in next iteration. As you
said
I will add

if (genParticles.isValid()){

}

PR #68, opened 2015-10-05
BTag SF + PU Weighting

rpatelCERN:
Hey Kevin, 
Here is a pull request for including some modules and input files to compute the PU weights, PU systematic shift, and the BTag SF/Shifts.


Production/test/runMakeTreeFromMiniAOD_cfg.py
```diff
@@ -11,11 +11,12 @@
 outfile=parameters.value("outfile","test_run")
 
 # background estimations on by default
-lostlepton=parameters.value("lostlepton", True)
-hadtau=parameters.value("hadtau", True)
-doZinv=parameters.value("doZinv", True)
-QCD=parameters.value("QCD", True)
-
+lostlepton=parameters.value("lostlepton", False)
```
kpedro88, line 14: Please do not change defaults in the run config.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -111,11 +113,13 @@ def makeTreeFromMiniAOD(
     from TreeMaker.WeightProducer.getWeightProducer_cff import getWeightProducer
     process.WeightProducer = getWeightProducer(process.source.fileNames[0])
     process.WeightProducer.Lumi                       = cms.double(1) #default: 1 pb-1 (unit value)
-    process.WeightProducer.PU                         = cms.int32(0) # PU S10 3 for S10 2 for S7
-    process.WeightProducer.FileNamePUDataDistribution = cms.string("NONE")
+    process.WeightProducer.PU                         = cms.int32(4) # PU S10 3 for S10 2 for S7
```
kpedro88, line 116: Can you fix the comment on this line so it makes sense with our current procedure?

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -642,7 +646,29 @@ def makeTreeFromMiniAOD(
                                          'JetsProperties:neutralHadronMultiplicity(Jets_neutralHadronMultiplicity)',
                                          'JetsProperties:photonMultiplicity(Jets_photonMultiplicity)',
                                          'JetsProperties:flavor(Jets_flavor)'])
-
+    if IncludeBEff:
+    	process.bTaggingEffAnalyzerAK5PF = cms.EDAnalyzer('BTaggingEffAnalyzer',
```
kpedro88: Why is this called AK5PF? We're using AK4PFCHS jets...

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -642,7 +646,29 @@ def makeTreeFromMiniAOD(
...
+        	PtNBins            = cms.int32(100),
+        	PtMin              = cms.double(0.),
+        	PtMax              = cms.double(1000.),
+        	EtaNBins           = cms.int32(60),
+        	EtaMin             = cms.double(-3.),
+        	EtaMax             = cms.double(3.)
+    	)
+   	process.Baseline += process.bTaggingEffAnalyzerAK5PF
+    if AddBtagSF:
+	process.BTagScale=cms.EDProducer('BTagScale',
```
kpedro88, line 663: Please convert all tab characters to 4 spaces, to be consistent with Python whitespace expectations.

WeightProducer/src/WeightProducer.cc
```diff
@@ -555,19 +621,37 @@ std::vector<double> WeightProducer::generateWeights(PUScenario sc, const TH1* da
       5.005E-06};
     npuProbs = npuSummer12_S10;
   }
-
+  if(sc==Spring15) nMaxPU=25;
   std::vector<double> result(nMaxPU);
+
   double s = 0.0;
+if(sc==Spring15){
```
kpedro88: Why is this here if it's all commented out?

rpatelCERN: Fixed everything below, but if there are other edits you recommend you can keep sending them my way. I will wait for a bit before making another pull request. 
-Rishi

Utils/src/BTaggingEffAnalyzer.cc
```diff
@@ -0,0 +1,233 @@
...
+  edm::Handle<int> isoMuonTracks;
+  edm::Handle<int> isoElectronTracks;
+  edm::Handle<bool> JetID; 
+
+  iEvent.getByLabel("HT", HT);
+  iEvent.getByLabel("MHT","Pt", MHT);
+  iEvent.getByLabel("DeltaPhi", "DeltaPhi1", DeltaPhi1);
+  iEvent.getByLabel("DeltaPhi","DeltaPhi2", DeltaPhi2);
+  iEvent.getByLabel("DeltaPhi","DeltaPhi3", DeltaPhi3);
+//  iEvent.getByLabel("DeltaPhi","DeltaPhi4", DeltaPhi4);
```
kpedro88: I think we do want to include DeltaPhi4 in the signal region selection from now on.

Utils/src/BTaggingEffAnalyzer.cc
```diff
@@ -0,0 +1,233 @@
...
+  iEvent.getByLabel("IsolatedMuonTracksVeto","isoTracks", isoMuonTracks);
+  iEvent.getByLabel("IsolatedElectronTracksVeto","isoTracks", isoElectronTracks);
+  iEvent.getByLabel("GoodJets", JetID);
+  bool passEvSel=false;
+  if(*HT>500 && *MHT>200 && *DeltaPhi1>0.5 && *DeltaPhi2>0.5 && *DeltaPhi3>0.3 && *NJets>=4 && *Leptons==0 && *isoPionTracks==0 && *isoMuonTracks==0 && *isoElectronTracks==0 && *JetID)passEvSel=true;
+ 
+  // loop over jets
+  for(PatJetCollection::const_iterator it = jets->begin(); it != jets->end(); ++it)
+  {
+    if(!passEvSel)continue;
```
kpedro88: It's inefficient to continue NJets times. Instead, just skip the whole for loop if the event fails selection. E.g., before the for loop, just put:

```
if(!passEvSel) return;
```

Utils/src/BTaggingEffAnalyzer.cc
rpatelCERN: Yes that is commented out in my version because I forked the Run2 branch before that variable was added. 
-Rishi

Utils/src/BTaggingEffAnalyzer.cc
```diff
@@ -0,0 +1,233 @@
...
+//  iEvent.getByLabel("DeltaPhi","DeltaPhi4", DeltaPhi4);
+
+  iEvent.getByLabel("NJets", NJets);
+  iEvent.getByLabel("LeptonsNew", Leptons);
+  iEvent.getByLabel("IsolatedPionTracksVeto","isoTracks", isoPionTracks);
+  iEvent.getByLabel("IsolatedMuonTracksVeto","isoTracks", isoMuonTracks);
+  iEvent.getByLabel("IsolatedElectronTracksVeto","isoTracks", isoElectronTracks);
+  iEvent.getByLabel("GoodJets", JetID);
+  bool passEvSel=false;
+  if(*HT>500 && *MHT>200 && *DeltaPhi1>0.5 && *DeltaPhi2>0.5 && *DeltaPhi3>0.3 && *NJets>=4 && *Leptons==0 && *isoPionTracks==0 && *isoMuonTracks==0 && *isoElectronTracks==0 && *JetID)passEvSel=true;
```
kpedro88: I'm not really thrilled about implementing the whole RA2/b-specific signal region selection in a TreeMaker module. Is it absolutely necessary for this workflow, or could the b-tagging efficiency analyzer be kept separate from TreeMaker somehow?

Utils/src/BTagScale.cc
```diff
@@ -0,0 +1,392 @@
+// -*- C++ -*-
```
kpedro88, line 1: The wildly inconsistent indentation in this file makes it difficult to read. Please revise if you have the time.

Utils/src/BTaggingEffAnalyzer.cc
rpatelCERN: I was thinking you would run the efficiency plots whenever you run over the full signal sample for the first time. Then the histograms are stored in an original set of trees and you don’t run this module again. 
-Rishi

Utils/src/BTagScale.cc
```diff
@@ -0,0 +1,392 @@
...
+        iEvent.put(b0down, "BTagProb0ShiftDown");
+        std::auto_ptr<double> b1down(new double(Prob1ShiftDown));
+        iEvent.put(b1down, "BTagProb1ShiftDown");
+        std::auto_ptr<double> b2down(new double(Prob2ShiftDown));
+        iEvent.put(b2down, "BTagProb2ShiftDown");
+        std::auto_ptr<double> b3down(new double(Prob3ShiftDown));
+        iEvent.put(b3down, "BTagProb3ShiftDown");
+
+
+	std::cout<<"BTagProb0 Nominal "<<Prob0<<" BTagProb1 "<<Prob1<<" BTag Prob2 "<<Prob2<<std::endl;
```
kpedro88: Is it necessary to print this message for every event?

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -642,7 +646,29 @@ def makeTreeFromMiniAOD(
...
+        	PtMax              = cms.double(1000.),
+        	EtaNBins           = cms.int32(60),
+        	EtaMin             = cms.double(-3.),
+        	EtaMax             = cms.double(3.)
+    	)
+   	process.Baseline += process.bTaggingEff
+    if AddBtagSF:
+	process.BTagScale=cms.EDProducer('BTagScale',
+		JetsTag            = cms.InputTag('HTJets'),
+		BTagEffInput   =cms.string('./data/test_run_RA2AnalysisTree.root'),
```
kpedro88, line 665: Do we need to track this ROOT file "test_run_RA2AnalysisTree.root"? If so, could it be given a more descriptive name?

rpatelCERN, line 665: For now that file is a test file that is also part of the pull request. The file-naming convention will follow the treemaker convention for the signal scan: Just need to input Mg, mLSP
-Rishi

Utils/src/BTaggingEffAnalyzer.cc
kpedro88: You should always make sure to synchronize with the main branch before submitting a PR. This PR is noted to have merge conflicts, so you need to rebase. See Step 7 here for the basics: https://gist.github.com/kpedro88/81a15ccb00d9491b7d41

rpatelCERN: Okay will rebase and make a new request. I also incorporated the recommended changes. 
Thanks, 
Rishi

Utils/src/BTaggingEffAnalyzer.cc
kpedro88: My concern is that, conceptually, the goal of TreeMaker is flattening ntuples, applying object IDs, and calculating related quantities from identified objects. Applying event selections and running analyzers is intended to happen downstream. Since we already store the b-tag discriminator value and parton flavor for all the GoodJets, I think this entire procedure fits better into user analysis code outside of TreeMaker... for example, if you use my signal skims, the duplication of the signal selection can be entirely avoided, and you just have to apply a few cuts on the jet collections. I could even write this into my analysis code pretty easily, to keep things organized.

Utils/src/BTagScale.cc
```diff
@@ -0,0 +1,392 @@
...
+  produces<double>("BTagProb0ShiftUp");
+  produces<double>("BTagProb1ShiftUp");
+  produces<double>("BTagProb2ShiftUp");
+  produces<double>("BTagProb3ShiftUp");	
+
+  produces<double>("BTagProb0ShiftDown");
+  produces<double>("BTagProb1ShiftDown");
+  produces<double>("BTagProb2ShiftDown");
+  produces<double>("BTagProb3ShiftDown");
+
```
kpedro88, line 113: Could these products be 3 vectors instead of 12 doubles?

Utils/src/BTaggingEffAnalyzer.cc
rpatelCERN: I can understand your concern that the Treemaker is mainly for creating an analysis ntuple.

My idea here was mainly, that it is real pain to run over the full mass scan. The treemaker will do it at least once to make an ntuple without the baseline. So I run this module (recommended by the btag pog) together with our analysis ntuplizer to make a set of histograms which can be used at analysis level. 

The efficiency analyzer should run pretty fast. 

-Rishi

Utils/src/BTagScale.cc
rpatelCERN, line 113: Sure I don’t mind. I made them doubles because it made it easier for plotting the nominal and shifted b-tag shapes.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
kpedro88, line 665: I'm still a bit confused about why we need to track this file. Do I understand correctly that there will be one of these for every model point in the signal scan? As I mentioned in the comment on BTaggingEffAnalyzer, it seems like this process is more appropriately placed post-ntuple production...

These uncertainties will only be used in the datacards for the signal samples, right? I think we could just calculate them on the fly when we make datacards.

Utils/src/BTaggingEffAnalyzer.cc
kpedro88: You don't have to make a new pull request when you rebase. You may have to force push to origin because the rebase rewrites history:

```
git push -f origin
```

but then it will automatically update this PR.

Utils/src/BTaggingEffAnalyzer.cc
kpedro88: It's true that the full mass scan has a ton of files in it. But we're going to have to run over them anyway to make datacards for the limit setting...

Maybe you, @nhanvtran, and I should have a conversation about the best workflow for calculating signal systematics before we go further.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
rpatelCERN, line 665: Yes that is correct. There will be a histogram for every model point (Mg, mLSP), which is why I wanted to include it when you ntuplize the full mass scan for the signal.  
-Rishi

Utils/src/BTagScale.cc
kpedro88, line 113: I think it's just as simple to plot "BTagProbShiftUp[0]" in TTree::Draw(), and it makes the trees a lot less cluttered.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
kpedro88, line 665: To me, it makes a lot more sense to do this alongside or after my skimming step, which makes the files that get used for the datacards. This comes after TreeMaker and after splitting the signal scans into individual model points. I think we should avoid duplicating the splitting effort if possible. Is there an integration team meeting tomorrow where we could discuss this? @nhanvtran

rpatelCERN, line 665: Okay I will take a look at your signal skims for now, and create a version of this eff Analzyzer that runs on the treemaker skims. 
-Rishi

kpedro88:
@rpatelCERN - as we discussed, I'm going to close this one, and then you can resubmit another one with just the pileup weighting part.

PR #81, opened 2015-10-21
Added track collections without mT or isolation cuts for T&P, plottin…

jbradmil:
Should no longer need a separate set of T&P trees nor the doProcessesForLLTP.py script.


kpedro88:
Looks good. Not too many new branches, and the increase in the size of the trees is minimal.

PR #92, opened 2015-11-03
recluster cleaned pfCandidates collection for Zinv / minor changes to ZProducer

fojensen:
ZProducer related (but unrelated to the jet cleaning/reclustering)
1) added ZCandidates RecoCandVector to the TreeMaker output
2) removed Zp4 and ZNum variables - just use ZCandidates->at(i) or ZCandidates->size()

GoodJetsProducer related
1) added flags to skip not applying JetID to jets overlapping with leptons or photons (bandaid on a bandaid) when doing Zinv cleaning

doZinv.py related
1) added flow to recreate event variables based on reclustering a jet collection after lepton/photon removal - this is controlled by two flags which decide if you want to remove all the leptons in the event, GoodPhotons:BestPhoton, or both. e.g. doZinv_leptons=true doZinv_photon=false means remove all the identified leptons in the event but leave the goodPhotons:BestPhoton collection in the candidate list which is the input for the jet reclustering
2) still need to implement the cleaning in the isolated tracks producer (see commented out code in doZinv.py) which in turn maybe warrants recomputing MET (used in mT cut for isolated tracks)
3) The old variables (*clean) are currently left in for comparison/validation - the new variables are called *cleanv2

see plots of the comparisons of the two cleaning methods in GJets and DYJets MC:
http://www-hep.colorado.edu/~fjensen/02102015_cleaningPR/leptons/
http://www-hep.colorado.edu/~fjensen/02102015_cleaningPR/photon/

thank you

godspeed


kpedro88:
for future reference, this was included in #96

PR #123, opened 2016-03-03
Added Mt2 producer to Run2 branch of sbein/TreeMaker fork

sbein:
The new producer should exactly reproduce DavisMt2


kpedro88:
@sbein - Thanks for taking care of this! Just a few comments to make sure it fits into the workflow nicely:
1. The MetTag argument for the producer should use the METTag variable defined near the top of makeTree... (just before the JEC stuff). Updated JECs have to be propagated to the MET, so in that case the good MET object has a different tag than the default.
2. Can you remove the `#needs validation` comments?
3. I would prefer fully capitalized "MT2" for the output branch name.

sbein:
@kpedro88  np, Kevin! Let me know if the latest version looks ok.

PR #138, opened 2016-03-29
Removing QCD module

aminghiasi:
Removing these variables from QCD module:

minDeltaPhiNames
RJetDeltaPhi
RJetMinDeltaPhiEta24
RJetMinDeltaPhiEta5
RJetMinDeltaPhiIndexEta24
RJetMinDeltaPhiIndexEta5
GenDeltaPhi
GenMinDeltaPhiEta24
GenMinDeltaPhiEta5
GenMinDeltaPhiIndexEta24
GenMinDeltaPhiIndexEta5


kpedro88:
Hi @aminghiasi,

Thanks for cutting out all the unneeded branches, this is very helpful. A few comments on the PR:
1. In general, if you are not using parts of the code anymore (python or c++), please delete the unused sections instead of just commenting them out. We can always go look at the history if you need to use something again, and it prevents us from being confused in the future when we forget why it was commented out.
2. Since we no longer need GenJets_invisibleEnergy, the whole process.GenJetProperties section in DeltaPhiQCD.py can be removed.

A few comments not directly related to the PR:
3. I discussed with Owen at the workshop today, and it would be great if you could check the GenParticles collections in the V6 ntuples to see if they have all the information you currently store in NeutrinoLorentzVector and related branches. If some of the information is missing, please let us know. Otherwise, we can reduce the number of branches even further. But again, I will accept this PR once changes 1 and 2 are made. We can do the gen neutrino check later.
4. At some point I might rearrange the GenJet stuff a little (hopefully to accommodate some R+S stuff in TreeMaker as well), but you don't have to worry about this.

-Kevin

PR #165, opened 2016-07-05
Added bool for bad muon jet event filter

jbradmil:


kpedro88:
@jbradmil 

In the past I've been reluctant to add these very contingent filters to TreeMaker. My expectation is that I'll launch ntuple production for the third time in three days, and then tomorrow someone will say "let's change the pT cut to 250!". It also seems a bit incongruous with our analysis to use DeltaPhi(jet,MET) rather than DeltaPhi(jet,MHT). But physics and forethought clearly aren't priorities for us anymore, so whatever.

If you're going to shove this into GoodJets, the default value `METTag="slimmedMETs"` is not acceptable. You need to update _all_ the `makeJetVars` calls, including those in doZinvBkg, using the proper MET tag.

PR #183, opened 2016-09-07
add minDeltaR variable for Zinv

fojensen:
add EDProducer to calculate minimum deltaR between MadGraph photon and partons.
see:
http://www-hep.colorado.edu/~fjensen/02092016_deltaR/
It is useful for making the GJets and GJets0p4 MC samples orthogonal
In the case when running on DY or Zinv MC the Z boson will take the place of the photon

### DIFF
--- a/TreeMaker/python/doZinvBkg.py
+++ b/TreeMaker/python/doZinvBkg.py
@@ -181,6 +181,11 @@ def doZinvBkg(process,tagname,geninfo,residual):
     process.TreeMaker2.VectorBool.append("goodPhotons:nonPrompt(photon_nonPrompt)")
     process.TreeMaker2.VectorBool.append("goodPhotons:fullID(photon_fullID)")
 
+    ## add MadGraph-level deltaR between photon or Z and partons
+    ## mainly used to make GJets and GJets0p4 MC samples orthogonal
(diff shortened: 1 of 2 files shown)


TreeMaker/python/doZinvBkg.py
```diff
@@ -181,6 +181,11 @@ def doZinvBkg(process,tagname,geninfo,residual):
     process.TreeMaker2.VectorBool.append("goodPhotons:nonPrompt(photon_nonPrompt)")
     process.TreeMaker2.VectorBool.append("goodPhotons:fullID(photon_fullID)")
 
+    ## add MadGraph-level deltaR between photon or Z and partons
+    ## mainly used to make GJets and GJets0p4 MC samples orthogonal
+    process.minDeltaR = cms.EDProducer("MinDeltaRDouble")
+    process.TreeMaker2.VarsDouble.extend(['minDeltaR'])
```
kpedro88: Can you give this branch a more descriptive name? "minDeltaR" is very generic.

At least "madMinDeltaR" would be good (our other MadGraph-based variable is madHT).

fojensen: I changed the name, I used the web-browser... hopefully that went through correctly

PR #187, opened 2016-10-13
Master

sbein:
I've created a new branch, so that the pull request will be independent from my previous pull request. The only development is the addition of an EDAnalyzer that is run when geninfo is flagged that pulls the pMSSM ID from the LHEEventProduct.


kpedro88:
We do not use the master branch. Closing this, you should make your changes on top of the Run2 branch.

PR #189, opened 2016-10-14
Added producer to save the pMSSM ID

sbein:
One new branch, PmssmId, is created and filled when the flag geninfo is set to true.

### DIFF
--- a/TreeMaker/python/makeTreeFromMiniAOD_cff.py
+++ b/TreeMaker/python/makeTreeFromMiniAOD_cff.py
@@ -107,11 +107,10 @@ pmssm=False
         )
         VarsDouble.extend(['SusyScan:SusyMotherMass','SusyScan:SusyLSPMass'])
 
-    if pmssm:
         # pMSSM ID for identifying the pMSSM model point
         from TreeMaker.Utils.pmssm_cfi import PmssmProducer
         process.Pmssm = PmssmProducer.clone(
-            shouldScan = cms.bool(fastsim),
+            shouldScan = cms.bool(pmssm),
             debug = cms.bool(False)
         )
         VectorString.extend(['Pmssm:PmssmId'])
--- a/TreeMaker/python/makeTreeFromMiniAOD_cff.py
+++ b/TreeMaker/python/makeTreeFromMiniAOD_cff.py
@@ -105,6 +105,16 @@ signal=False
             isLHE = cms.bool(False)
         )
         VarsDouble.extend(['SusyScan:SusyMotherMass','SusyScan:SusyLSPMass'])
+
+        # pMSSM ID for identifying the pMSSM model point
+        from TreeMaker.Utils.pmssm_cfi import PmssmProducer
+        process.Pmssm = PmssmProducer.clone(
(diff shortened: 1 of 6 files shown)


TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -105,6 +105,16 @@ def makeTreeFromMiniAOD(
             isLHE = cms.bool(False)
         )
         VarsDouble.extend(['SusyScan:SusyMotherMass','SusyScan:SusyLSPMass'])
+
+        # pMSSM ID for identifying the pMSSM model point
+        from TreeMaker.Utils.pmssm_cfi import PmssmProducer
+        process.Pmssm = PmssmProducer.clone(
+            shouldScan = cms.bool(fastsim),
```
kpedro88: I don't think we want to run this for all fastsim. Maybe we should add a separate pmssm parameter to makeTree and runMakeTree.

Utils/src/PmssmProducer.cc
```diff
@@ -0,0 +1,151 @@
...
+				if(comment.back()=='\n') comment.pop_back();
+				if(debug_) std::cout << comment << std::endl;
+				pmssmId.push_back(comment); 
+				pmssmXsec = 0.0;
+				break;//finished with this event
+			}
+		}
+	}
+
+	std::auto_ptr<std::vector<std::string> > pmssmId_(new std::vector<std::string>(pmssmId));
```
kpedro88: please use `std::unique_ptr` and `std::move` for products

kpedro88: (see e.g. the latest version of https://github.com/TreeMaker/TreeMaker/blob/Run2/Utils/src/SusyScanProducer.cc)

kpedro88:
thanks for catching it in the other producer too

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -105,6 +106,17 @@ def makeTreeFromMiniAOD(
             isLHE = cms.bool(False)
         )
         VarsDouble.extend(['SusyScan:SusyMotherMass','SusyScan:SusyLSPMass'])
+
+    if pmssm:
+        # pMSSM ID for identifying the pMSSM model point
+        from TreeMaker.Utils.pmssm_cfi import PmssmProducer
+        process.Pmssm = PmssmProducer.clone(
+            shouldScan = cms.bool(fastsim),
```
kpedro88: The idea was to keep the pMSSM branches in all MC ntuples (to avoid unnecessary digressions in the ntuple format), but only fill them when the pmssm parameter is activated, i.e.:

```
shouldScan = cms.bool(pmssm)
```

and remove the `if pmssm` above.

kpedro88:
Is there any pMSSM sample file from 80X that we can use to test this yet?

sbein:
Not for a couple weeks, but I used this code (successfully) with 74x samples some time ago.

kpedro88:
Okay, I will merge now and we can tweak later if needed.

PR #195, opened 2016-10-26
make MinDeltaRDouble more versatile

fojensen:
two modifications:

1) instead of selecting an isHardProcess() photon or Z, use just the status 23 objects (which is a subset, in the photon/Z MC there is only ever 1 status 23 boson)

2) add functionality such that if there is no status 23 photon or Z in the event, minDeltaR will be calculated with respect to the highest pT photon of any status (which in the case of a QCD test file I looked at is always status 1)


kpedro88:
Note for myself: I tried to think of a way to combine the first two loops over GenParticles into one loop, but in the end I don't think it would improve performance at all.

PR #197, opened 2016-10-28
make MinDeltaRDouble useful for fragmentation, fix Z status bug

fojensen:
added a status int to allow what sort of photon or Z is chosen to calculate deltaR with - for example, for fragmentation we would like to use photons that are radiated from quarks, not from meson decay

fix bug where I initially assumed the hard process Z had status 23, it in fact has status 22

for expected output on MC samples relevant to Zinv, see:
http://www-hep.colorado.edu/~fjensen/28102016_deltaRVerify/


fojensen:
(maybe this is useful) if the status is:
23 -> a hard process Z was found (Zinv/DY MC)
22 -> a hard process photon was found (GJets MC)
1 -> a photon with a quark mother was found (QCD MC)
-1 -> a photon with any mother was found
0 -> the search failed and minDR=999.

fojensen:
for more updated plots containing expected output see:
http://www-hep.colorado.edu/~fjensen/30102016_deltaRStatusWCuts/

PR #203, opened 2016-12-09
Made updates necessary for processing pMSSM samples. 

sbein:
The main remaining update, besides possible consistency tweaks you say are needed, is the pMSSM cross section file, which will be updated in the coming days. 
-Changed the PmssmId variable (branch) to a double
-Added cross section map text file for scan-1 pMSSM points
-Modified a few lines in getWeightProducer_cff.py to trigger pmssm when appropriate. This swaps out SusyScan:SusyMotherMass for Pmssm:PmssmId as the key used to identify the cross section.
-Used get_py.py script to produce pMSSM file lists and added them to Production/python/Spring16Fast
-Added an entry to unitTest.py for the pMSSM dataset


Utils/src/PmssmProducer.cc
```diff
@@ -15,80 +15,109 @@
...
 		
-		void process(std::string line, char delim, std::vector<std::string>& fields);
+  void process(std::string line, char delim, std::vector<std::string>& fields);
+  double getXsec(double pmssmid);
 		
-		// ----------member data ---------------------------
-		edm::GetterOfProducts<LHEEventProduct> getterOfProducts_;
-		bool shouldScan_, debug_;
+  // ----------member data ---------------------------
+  edm::GetterOfProducts<LHEEventProduct> getterOfProducts_;
```
kpedro88, line 43: Are the pMSSM samples still produced with the old-style LHE method, i.e. this producer doesn't need to access `GenLumiInfoHeader`?

Utils/src/PmssmProducer.cc
```diff
@@ -15,80 +15,109 @@
...
+  // (e.g. close files, deallocate resources etc.)
+}
+
+double PmssmProducer::getXsec(double pmssmid)
+{
+  double xsec = -1;
+  std::ifstream fileInput;
+  std::string line;//use a sprintf with pmssmid
+  char search [30];
+  sprintf (search, "%d", int(pmssmid));
```
kpedro88: Please use `stringstream` and regular `std::string` instead of char arrays and `sprintf`

Utils/src/PmssmProducer.cc
```diff
@@ -15,80 +15,109 @@
...
+  fileInput.open(_xsecFilename.c_str());
+  if(fileInput.is_open()){
+    while(getline(fileInput, line)) {
+      if (size_t(line.find(search, 0)) != std::string::npos) {
+	xsec = atof(line.substr(line.find("\t"),line.size()).c_str());
```
kpedro88: Please use `stringstream` instead of `atof`

Utils/src/PmssmProducer.cc
```diff
@@ -15,80 +15,109 @@
...
 {
 	
-	// do anything here that needs to be done at desctruction time
-	// (e.g. close files, deallocate resources etc.)
-	
+  // do anything here that needs to be done at desctruction time
+  // (e.g. close files, deallocate resources etc.)
+}
+
+double PmssmProducer::getXsec(double pmssmid)
```
kpedro88: Reading the entire xsec file should be done once, in the constructor, to create a hash table that can be accessed whenever an xsec is needed. Doing this every single event is very inefficient. (See example in `WeightProducer`.)

Utils/src/PmssmProducer.cc
```diff
@@ -15,80 +15,109 @@
...
+    edm::Handle<LHEEventProduct> product = handles[0];
+    for(LHEEventProduct::comments_const_iterator cit = product->comments_begin(); cit != product->comments_end(); ++cit){
+      size_t found = (*cit).find(".slha");
+      if(found != std::string::npos){
+	std::string comment = (*cit).substr(1,(*cit).find(".slha")-1);
+	if(comment.back()=='\n') comment.pop_back();
+	std::vector<std::string> nameblocks;
+	process(comment, '_', nameblocks);
+	char character_string [30];
+	sprintf (character_string, "%s%s", nameblocks[2].c_str(),nameblocks[3].c_str());
```
kpedro88: Please use string concatenation instead of char arrays and `sprintf`, i.e.:
```
string character_string = nameblocks[2]+nameblocks[3]
```

WeightProducer/src/WeightProducer.cc
```diff
@@ -90,23 +90,21 @@ WeightProducer::WeightProducer(const edm::ParameterSet& iConfig) :
    _weightName(iConfig.getParameter<edm::InputTag> ("weightName")),
    _genEvtTag(edm::InputTag("generator")),
    _puInfoTag(edm::InputTag("slimmedAddPileupInfo")),
-   _SusyMotherTag(edm::InputTag("SusyScan:SusyMotherMass")),
+   ////_SusyMotherTag(edm::InputTag("SusyScan:SusyMotherMass")),
```
kpedro88: remove unneeded line instead of commenting out

WeightProducer/src/WeightProducer.cc
```diff
@@ -117,15 +115,14 @@ WeightProducer::WeightProducer(const edm::ParameterSet& iConfig) :
...
-   
    // Option 4: assign cross section for FastSim
    else if (_weightingMethodName == "FastSim") {
       _weightingMethod = FastSim;
       std::cout << "WeightProducer: Finding cross sections for FastSim" << std::endl;
       std::cout << "  Target luminosity (1/pb) : " << _lumi << std::endl;
       std::cout << "          Number of events : " << _NumberEvents << std::endl;
       _SusyMotherTok = consumes<double>(_SusyMotherTag);
-     
+      std::cout << "entered the weight producer E" << std::endl;
```
kpedro88: is this cout necessary?

kpedro88:
Done with code review, looks good except for the few changes/clarifications requested

WeightProducer/src/WeightProducer.cc
sbein: Struck line

WeightProducer/src/WeightProducer.cc
sbein: Struck commented line

Utils/src/PmssmProducer.cc
sbein: Ok this is done now

Utils/src/PmssmProducer.cc
sbein: Yep, the pMSSM ID is embedded in the LHEEventProduct.

Utils/src/PmssmProducer.cc
sbein: This function is never called in TreeMaker because the config boolean "do_xec" is set to false. This function is not called in TreeMaker because I modified the existing WeightProducer modules to get the cross section to be seamless with the way the other FastSim xsecs are obtained (the map is read out once by weightproducer, etc, by the existing function). The function referred to is only there because it is part of the  official EDProducer being passed around to the SUSY teams doing a pMSSM interpretation.

Utils/src/PmssmProducer.cc
sbein: Again, hopefully since this function is not called by TreeMaker, you are fine having that sitting there.

Utils/src/PmssmProducer.cc
sbein: Agreed, but again, this doesn't actually get called in TreeMaker.

kpedro88:
I would prefer not to merge inefficient/outdated code in TreeMaker just because it's what the SUSY group is using. I am sure everyone would appreciate having a modern, efficient producer to get pMSSM info.

PR #207, opened 2017-01-05
Btag updates

kpedro88:
Updated WPs according to https://twiki.cern.ch/twiki/bin/viewauth/CMS/BtagRecommendation80XReReco and double b-tagger training according to https://twiki.cern.ch/twiki/bin/view/CMS/Hbbtagging#V4_training. Also updated CMSSW version to get latest xrootd version (4.5.0).


awhitbeck:
Thanks!

PR #208, opened 2017-01-18
Updated prompt photons and added 2 new branches

vhegde91:
Prompt photon definitions have been updated. A new branch is added to check whether photon is real or electron faking. Also added a branch to check if there is any hard scattered photon from quarks/leptons/bosons using only gen level information.
//by Vinay

### DIFF
--- a/Utils/src/PhotonIDisoProducer.cc
+++ b/Utils/src/PhotonIDisoProducer.cc
@@ -77,6 +77,7 @@ PhotonIDisoProducer::PhotonIDisoProducer(const edm::ParameterSet& iConfig):
   produces< std::vector< bool > >("hadronization");
   produces< std::vector< bool > >("nonPrompt");
   produces< std::vector< bool > >("fullID");
+  produces< std::vector< bool > >("electronFakes");
 }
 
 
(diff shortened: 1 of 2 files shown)


Utils/src/PhotonIDisoProducer.cc
```diff
@@ -254,22 +257,33 @@ PhotonIDisoProducer::produce(edm::Event& iEvent, const edm::EventSetup& iSetup)
...
+		if( gen.Pt()/photon.Pt() > 0.5 && gen.Pt()/photon.Pt() < 1.5 )
+		  matchedGenPrompt++ ;
+	      }//for prmopt photons
+            }//gen matching
           }
+	  //check wheather photon has matched to a gen electron or not
+	  if( abs(iGen->pdgId()) == 11 && iGen->status() == 1 ){
+	    TLorentzVector genE( iGen->px() , iGen->py() , iGen->pz() , iGen->energy() );
+            TLorentzVector photon( iPhoton->px() , iPhoton->py() , iPhoton->pz() , iPhoton->energy() );
+	    if( genE.DeltaR(photon) < 0.2 ){
```
kpedro88, line 283: It is computationally somewhat wasteful to create TLorentzVectors here just to calculate deltaR. I recommend this approach instead:
https://github.com/TreeMaker/TreeMaker/blob/Run2/Utils/src/MinDeltaRDouble.cc#L11
https://github.com/TreeMaker/TreeMaker/blob/Run2/Utils/src/MinDeltaRDouble.cc#L84

All Candidate objects have a `p4()` method.

kpedro88: Actually never mind, I noticed this is done in other places that you didn't introduce, so I'll just fix it later

PR #210, opened 2017-01-19
change to PUPPI AK8 jets

awhitbeck:
…ts and their corresponding substructure properties.  Main TreeMaker configuration changed so that old CHS AK8 jet properties are replaced with new puppi AK8 jet properties.  Jet mass is now the softdrop mass instead of the pruned jet mass.  Minor change to makeJetVars so that user can opt out of automatically getting the vector<TLorentzVector> branch for the corresponding jet collection.


awhitbeck:
closing PR to rework code based on new information from Jet/MET experts.

PR #213, opened 2017-01-20
fixed deltaR,updated prmopt photon and added new branches Photons_ele…

vhegde91:
used p4() instead of defining TLorentzVector


kpedro88:
closing this because I already started going through and fixing this (not just in this one instance, also elsewhere), but thanks anyway

PR #215, opened 2017-01-22
add lepton/photon cleaned AK8 jets

fojensen:
for this PR to work the following will need to be added to the TreeMaker installation recipe:
cd CMSSW_8_0_X/src/
git clone git@github.com:cms-jet/JetToolbox.git JMEAnalysis/JetToolbox -b jetToolbox_80X_V3

for other info reference the JetToolbox twiki:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/JetToolbox

the attached plots show distributions comparing the new electron/muon/photon cleaned AK8 jets to the 'standard' collection of AK8 jets - for events with no electrons, muons, or photons - the agreement is expected

[compareJets.doubleBDiscriminator.pdf](https://github.com/TreeMaker/TreeMaker/files/722244/compareJets.doubleBDiscriminator.pdf)
[compareJets.NJets8.pdf](https://github.com/TreeMaker/TreeMaker/files/722241/compareJets.NJets8.pdf)
[compareJets.pt.pdf](https://github.com/TreeMaker/TreeMaker/files/722242/compareJets.pt.pdf)
[compareJets.puppiSoftDropMass.pdf](https://github.com/TreeMaker/TreeMaker/files/722243/compareJets.puppiSoftDropMass.pdf)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ git cms-merge-topic -u cms-btv-pog:BoostedDoubleSVTaggerV4-WithWeightFiles-v1_fr
 git cms-merge-topic -u kpedro88:METfix8022
 git cms-merge-topic -u cms-met:fromCMSSW_8_0_20_postICHEPfilter
 git cms-merge-topic -u kpedro88:storeJERFactor8022
+git clone git@github.com:cms-jet/JetToolbox.git JMEAnalysis/JetToolbox -b jetToolbox_80X_V3
 git clone git@github.com:TreeMaker/TreeMaker.git -b Run2
 scram b -j 8
 cd TreeMaker/Production/test
--- a/TreeMaker/python/doZinvBkg.py
+++ b/TreeMaker/python/doZinvBkg.py
@@ -1,6 +1,85 @@
 import FWCore.ParameterSet.Config as cms
 
 def reclusterZinv(process, geninfo, residual, cleanedCandidates, suff, fastsim):
+    
+    ### AK8 detour
+
+    # https://twiki.cern.ch/CMS/JetToolbox
+    from JMEAnalysis.JetToolbox.jetToolbox_cff import jetToolbox
+    listBTagInfos = ['pfInclusiveSecondaryVertexFinderTagInfos'] 
+    listBtagDiscriminatorsAK8 = [
+        'pfCombinedInclusiveSecondaryVertexV2BJetTags',
+        'pfBoostedDoubleSecondaryVertexAK8BJetTags'
+    ]
+    jetToolbox(process,
+        'ak8',
+        'jetSequence',
+        'out',
+        PUMethod = 'Puppi',
+        miniAOD = True,
+        runOnMC = geninfo,
+        newPFCollection = True,
+        nameNewPFCollection = 'cleanedCandidates',
+        Cut = 'pt>170.',
+        addSoftDrop = True,
+        addSoftDropSubjets = True,
+        addNsub = True,
+        bTagInfos = listBTagInfos, 
+        bTagDiscriminators = listBtagDiscriminatorsAK8, 
+    )
+    JetAK8CleanTag = cms.InputTag("packedPatJetsAK8PFPuppiSoftDrop")
+
+    from TreeMaker.TreeMaker.makeJetVars import makeJetVars
+    makeJetVars(process,
+        JetTag=JetAK8CleanTag,
+        suff='AK8Clean',
+        skipGoodJets=False,
+        storeProperties=1,
+        geninfo=geninfo,
+        fastsim=fastsim,
+        onlyGoodJets=True
+    )
+    
+    from TreeMaker.Utils.jetproperties_cfi import jetproperties
+    process.JetsPropertiesAK8Clean = jetproperties.clone(
+        JetTag       = JetAK8CleanTag,
+        debug = cms.bool(False),
+        properties = cms.vstring(
+            "PuppiSoftDropMass"    ,
+            "NsubjettinessTau1"    ,
+            "NsubjettinessTau2"    ,
+            "NsubjettinessTau3"    ,
+            "bDiscriminatorSubjet1",
+            "bDiscriminatorSubjet2",
+            "bDiscriminatorCSV"    ,
+            "NumBhadrons"          ,
+            "NumChadrons"          ,
+        )
+    )
+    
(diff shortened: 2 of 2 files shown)


fojensen:
the referenced plots were made with 10000 events from the T1tttFast MC sample:
https://github.com/TreeMaker/TreeMaker/blob/Run2/Production/test/unitTest.py#L50
requiring:
@Electrons->size()==0 && @Muons->size()==0 && @Photons->size()==0

kpedro88:
A couple comments before I actually test it myself:
1) Can you add the new recipe lines to the README?
2) Will it be acceptable for the Higgs+MET analysis to use just PUPPI for the cleaned jets even though the regular AK8 is still CHS? @awhitbeck

fojensen:
yes - of course I can add the line to the README file. I am currently
looking around to see if I can somehow modify the PR through the web
browser...

but it will amount to adding
git clone git@github.com:cms-jet/JetToolbox.git JMEAnalysis/JetToolbox -b
jetToolbox_80X_V3
in between the following two lines

git cms-merge-topic -u kpedro88:storeJERFactor8022
git clone git@github.com:TreeMaker/TreeMaker.git -b Run2

stay tuned...

kpedro88:
If you go to your fork/branch on github, you can edit the file in the web browser.

fojensen:
In response to Question 2...

I tried to use the same variables as what are used in the main part of
the code (i.e. in makeTreeFromMiniAOD_cff.py). A recent PR by
@awhitbeck:
https://github.com/TreeMaker/TreeMaker/pull/211
switched to Puppi as well...

kpedro88:
That one only uses PUPPI for the soft drop mass and nsubjettiness variables. The jet fourvectors and btag discriminants still come from CHS, because the PUPPI jet info wasn't stored in an easily usable way. I guess we should think more about this. I would prefer to avoid adding even more reclusterings...

fojensen:
I see what you mean... by looking at the standard MiniAOD includes,
the slimmedJetsAK8 collection is CHS, but the
slimmedJetsAK8PFPuppiSoftDropPacked collection is of subjets using the
Puppi algorithm:
https://twiki.cern.ch/CMSPublic/WorkBookMiniAOD2016

If that is what is 'officially' supported or generally used, perhaps I
should change the newly reclustered jets to use CHS instead of Puppi?
But then again, I should look more in to how Puppi works...

awhitbeck:
I think its not a big deal to have PUPPI for the main jet collection.  I was told that typically CHS is used for everything other than substructure variables.  However, I also don't think there is an official recommendation.  I'll have to chat more with the experts on this.  

@fojensen can you remake your comparison of the jet mass after a pt cut, if you didn't already?  The other variables seem to agree quite well, I am just wondering why the masses are different.  Is it the PUPPI softdrop mass in both cases?

kpedro88:
I ran igprof w/out and w/ this PR. The addition in this PR makes TreeMaker take 2x longer per event (using GJets HT 600toInf). The primary slowdown seems to come from rerunning PUPPI on the cleaned PFCandidate collection. There are already PUPPI weights for the full PFCandidate collection stored in miniAOD, but I think the jet toolbox is rerunning PUPPI because we told it there was a new PFCandidate collection. We should try to figure out how to avoid rerunning PUPPI in the jet toolbox if possible (since the new collection is a subset of the old one, I would hope we could reuse the old weights...)

We can also think about reclustering to get AK8 PUPPI jets for the main collection in order to have consistent collections, but we have to solve this CPU issue first.

fojensen:
are the NJettiness variables used at all?
there are 6 branches with those variables in the TreeMaker. I also think
that JetToolbox has to do additional stuff to compute those variables

kpedro88:
There is a separate NjettinessAdder, but this is overall a pretty small amount of CPU.

I found this condition in the jet toolbox:
```
                        if ('puppi' in tmpPfCandName): 
                                srcForPFJets = pfCand
                                print '|---- jetToolBox: Not running puppi algorithm because keyword puppi was specified in nameNewPFCollection, but applying puppi corrections.' 
```
so I'll try renaming the collection and see if it works...

fojensen:
@awhitbeck
- the jets have a pt cut of 170 GeV, as is standard for the slimmedJetAK8
collection...
- yes, the mass distributions should be identical and be the 'puppi soft
drop jet mass', they look similar to me but it looks like the 'Clean'
version of the variable simply has more zero-jet masses

kpedro88:
I sent an email to the JetMET guys trying to figure out why the CPU usage is behaving strangely.

Also, since apparently the softdrop calculations require their own effective reclustering, it might be actually more efficient to do another PFCHS reclustering in the Zinv estimation (in addition to the PUPPI reclustering introduced in this PR), rather than doing another PUPPI reclustering for the main collection. But then we would have to replicate the "merging" process in https://github.com/cms-sw/cmssw/pull/13796, and I'm not sure about the CPU impact of that...

@fojensen, another thing needed in this PR is to set the jet corr levels manually, as we do for the AK4 reclustering:
```
    jecLevels = ['L2Relative', 'L3Absolute']
    if residual: jecLevels.append("L2L3Residual")
    jetToolbox(
    ...
    JETCorrLevels = jecLevels,
    subJETCorrLevels = jecLevels,
    )
```
The jet toolbox automatically turns residuals on if runOnMC is false, but to be consistent with the rest of TreeMaker, I would prefer to control them manually using our specific residual flag. There are some cases where we might not want to apply residuals to data (if they're missing or inaccurate).

I also found that setting `maxTau=3` for `jetToolbox` reduces the CPU usage a bit (the default is 4, but we don't use tau4).

fojensen:
I made ratio plots out of the previously shown variables. I additionally now include plots of the jet eta and jet phi
[compareJets.doubleBDiscriminator.pdf](https://github.com/TreeMaker/TreeMaker/files/727648/compareJets.doubleBDiscriminator.pdf)
[compareJets.eta.pdf](https://github.com/TreeMaker/TreeMaker/files/727645/compareJets.eta.pdf)
[compareJets.NJets8.pdf](https://github.com/TreeMaker/TreeMaker/files/727646/compareJets.NJets8.pdf)
[compareJets.phi.pdf](https://github.com/TreeMaker/TreeMaker/files/727650/compareJets.phi.pdf)
[compareJets.pt.pdf](https://github.com/TreeMaker/TreeMaker/files/727647/compareJets.pt.pdf)
[compareJets.puppiSoftDropMass.pdf](https://github.com/TreeMaker/TreeMaker/files/727649/compareJets.puppiSoftDropMass.pdf)





.

PR #217, opened 2017-01-23
Prepare for production with Summer16 MC, clean out old stuff

kpedro88:
This pull request is really just an attempt to see how many files I can add/remove at once.

NB, the Summer16 JECs assigned in the scenarios right now are definitely not final and should not be trusted.


kpedro88:
also NB, getWeightProducer notes a few missing MC samples

PR #244, opened 2017-09-15
Standardize and add variables for AK8 jets (including analysis-specific)

kpedro88:
This PR adds two producers to be used in a forthcoming search for semi-visible jets:
1. `BasicSubstructureProducer` - adds some simple substructure variables for AK8 jets.
2. `HiddenSectorProducer` - adds some jet/MET-related quantities (masses, DeltaPhi) specifically for the semi-visible jet search.

The two new producers can be activated with the `semivisible` flag (off by default).

This PR also:
* adds hidden sector particle IDs to the list of gen particles stored in the ntuples
* adds a flag `systematics` (true by default) that controls the JEC/JER variations for jets
* contains a few commits following up on #243 (reorganizing Python classes/interfaces).

Todo: make a version of `makeJetVars` to organize all AK8 jet variables (reduce code duplication for Zinv).


kpedro88:
Updated title to reflect growth of this PR:
* standardized jet properties function `makeJetVarsAK8` (used for normal and Zinv)
* add softdrop mass
* add subjets (requested for stealth SUSY analysis)
* add soft drop subjets for Zinv (using CHS, not puppi - will switch to puppi once it is the miniAOD default)
* standardize branch names for quark-gluon tagging input variables
* removed subjet b-tag discriminators (not used)
* proper handling of AK8 constituents (stored in a weird way in miniAOD, see https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookMiniAOD2016#Jets)

PR #256, opened 2017-12-08
save HLT objects associated with HLT_Ele27_WPTight_Gsf_v and pt>25

vhegde91:


Utils/src/TriggerProducer.cc
```diff
@@ -48,6 +50,9 @@ class TriggerProducer : public edm::global::EDProducer<> {
   edm::EDGetTokenT<edm::TriggerResults> trigResultsTok_;
   edm::EDGetTokenT<pat::PackedTriggerPrescales> trigPrescalesTok_;
   std::vector<std::string> parsedTrigNamesVec;
+  edm::EDGetTokenT<edm::TriggerResults> trigResultsToken;
```
kpedro88: we already have `trigResultsTok_`, don't need another one

Utils/src/TriggerProducer.cc
```diff
@@ -89,9 +94,15 @@ TriggerProducer::TriggerProducer(const edm::ParameterSet& iConfig)
   trigResultsTok_ = consumes<edm::TriggerResults>(trigResultsTag_);
   trigPrescalesTok_ = consumes<pat::PackedTriggerPrescales>(trigPrescalesTag_);
 
+  edm::InputTag theTriggerLabel("TriggerResults", "", "HLT");
```
kpedro88: remove

Utils/src/TriggerProducer.cc
```diff
@@ -89,9 +94,15 @@ TriggerProducer::TriggerProducer(const edm::ParameterSet& iConfig)
...
+  edm::InputTag theTrigObjLabel("selectedPatTrigger");
```
kpedro88: make configurable

Utils/src/TriggerProducer.cc
```diff
@@ -89,9 +94,15 @@ TriggerProducer::TriggerProducer(const edm::ParameterSet& iConfig)
...
+  trigResultsToken = consumes<edm::TriggerResults>(theTriggerLabel);
```
kpedro88: remove

Utils/src/TriggerProducer.cc
```diff
@@ -89,9 +94,15 @@ TriggerProducer::TriggerProducer(const edm::ParameterSet& iConfig)
...
+  trigObjCollToken = consumes<pat::TriggerObjectStandAloneCollection>(theTrigObjLabel);
+
   produces<std::vector<std::string> >("TriggerNames");
   produces<std::vector<int> >("TriggerPass");
   produces<std::vector<int> >("TriggerPrescales");
+  produces<std::vector<TLorentzVector> >("HLTElectronObjects");
```
kpedro88: add a configurable bool so it doesn't produce this all the time, only when requested

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +162,36 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
     }
   }
 
+  edm::Handle<edm::TriggerResults> triggerBits;
+  iEvent.getByToken(trigResultsToken,triggerBits);
```
kpedro88: remove these two lines

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +162,36 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
...
+
+  const edm::TriggerNames& names = iEvent.triggerNames(*triggerBits);
```
kpedro88: remove this, use `trigNames` defined above

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +162,36 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
...
+  edm::Handle<pat::TriggerObjectStandAloneCollection> triggerObjects;
```
kpedro88: put all of this behind bool

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +162,36 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
...
+  iEvent.getByToken(trigObjCollToken,triggerObjects);
+  //save the trigger object corresponding to the trigger HLT_Ele27_WPTight_Gsf_v*. Obtained code from https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookMiniAOD2016#Trigger
+  // loop over selected trigger objects                                                                                              
+  for (pat::TriggerObjectStandAlone obj : *triggerObjects) {
+    obj.unpackPathNames(names);
+    if(obj.pt() < 25.0) continue;//look for HLT objects with Pt > 25GeV only.
```
kpedro88: do this before unpacking

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +162,36 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
...
+    TLorentzVector objVec(obj.px(),obj.py(),obj.pz(),obj.energy());
```
kpedro88: remove this line

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +162,36 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
...
+    std::vector<std::string> pathNamesAll = obj.pathNames(false);
+    std::vector<std::string> pathNamesLast = obj.pathNames(true);
+    for (unsigned h = 0, n = pathNamesAll.size(); h < n; ++h) {
+      bool isBoth = obj.hasPathName( pathNamesAll[h], true, true );//object is associated wih l3 filter and associated to the last filter of a successfull path. this object caused the trigger to fire.
+      std::string path_i = pathNamesAll[h];
+      path_i.pop_back(); path_i.pop_back(); //remove last 2 characters from path name.
+      if( isBoth && ( (path_i == "HLT_Ele27_WPTight_Gsf_v") || (path_i == "HLT_Ele27_WPTight_Gsf_") ) ){ //check if the trigger path name is "HLT_Ele27_WPTight_Gsf_vX" or "HLT_Ele27_WPTight_Gsf_vXX"
+	hltEleObj->push_back(objVec);//save the trigger object corresponding to HLT_Ele27_WPTight_Gsf_v* trigger.
```
kpedro88: use `emplace_back(obj.px(),obj.py(),obj.pz(),obj.energy())`

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +162,36 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
...
+  iEvent.getByToken(trigObjCollToken,triggerObjects);
+  //save the trigger object corresponding to the trigger HLT_Ele27_WPTight_Gsf_v*. Obtained code from https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookMiniAOD2016#Trigger
+  // loop over selected trigger objects                                                                                              
+  for (pat::TriggerObjectStandAlone obj : *triggerObjects) {
+    obj.unpackPathNames(names);
+    if(obj.pt() < 25.0) continue;//look for HLT objects with Pt > 25GeV only.
+    TLorentzVector objVec(obj.px(),obj.py(),obj.pz(),obj.energy());
+    std::vector<std::string> pathNamesAll = obj.pathNames(false);
+    std::vector<std::string> pathNamesLast = obj.pathNames(true);
+    for (unsigned h = 0, n = pathNamesAll.size(); h < n; ++h) {
```
kpedro88: `(unsigned h = 0; h < pathNamesAll.size(); ++h)`

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +162,36 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
...
+  const edm::TriggerNames& names = iEvent.triggerNames(*triggerBits);
+  edm::Handle<pat::TriggerObjectStandAloneCollection> triggerObjects;
+  iEvent.getByToken(trigObjCollToken,triggerObjects);
+  //save the trigger object corresponding to the trigger HLT_Ele27_WPTight_Gsf_v*. Obtained code from https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookMiniAOD2016#Trigger
+  // loop over selected trigger objects                                                                                              
+  for (pat::TriggerObjectStandAlone obj : *triggerObjects) {
+    obj.unpackPathNames(names);
+    if(obj.pt() < 25.0) continue;//look for HLT objects with Pt > 25GeV only.
+    TLorentzVector objVec(obj.px(),obj.py(),obj.pz(),obj.energy());
+    std::vector<std::string> pathNamesAll = obj.pathNames(false);
```
kpedro88: `const std::vector<std::string>&`

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +162,36 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
...
+    std::vector<std::string> pathNamesLast = obj.pathNames(true);
```
kpedro88: `const std::vector<std::string>&`

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +162,36 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
...
+    for (unsigned h = 0, n = pathNamesAll.size(); h < n; ++h) {
+      bool isBoth = obj.hasPathName( pathNamesAll[h], true, true );//object is associated wih l3 filter and associated to the last filter of a successfull path. this object caused the trigger to fire.
+      std::string path_i = pathNamesAll[h];
+      path_i.pop_back(); path_i.pop_back(); //remove last 2 characters from path name.
```
kpedro88: remove this and in the condition on the next line, just use: `path_i.find("HLT_Ele27_WPTight_Gsf_v")!= std::string::npos`

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +162,36 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
...
+  // loop over selected trigger objects                                                                                              
+  for (pat::TriggerObjectStandAlone obj : *triggerObjects) {
+    obj.unpackPathNames(names);
+    if(obj.pt() < 25.0) continue;//look for HLT objects with Pt > 25GeV only.
+    TLorentzVector objVec(obj.px(),obj.py(),obj.pz(),obj.energy());
+    std::vector<std::string> pathNamesAll = obj.pathNames(false);
+    std::vector<std::string> pathNamesLast = obj.pathNames(true);
+    for (unsigned h = 0, n = pathNamesAll.size(); h < n; ++h) {
+      bool isBoth = obj.hasPathName( pathNamesAll[h], true, true );//object is associated wih l3 filter and associated to the last filter of a successfull path. this object caused the trigger to fire.
+      std::string path_i = pathNamesAll[h];
```
kpedro88: `const std::string&`

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +162,36 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
...
+	hltEleObj->push_back(objVec);//save the trigger object corresponding to HLT_Ele27_WPTight_Gsf_v* trigger.
+	break;
+      }
+    }
+  }
+
   iEvent.put(std::move(passTrigVec),"TriggerPass");
   iEvent.put(std::move(trigPrescaleVec),"TriggerPrescales");
   iEvent.put(std::move(trigNamesVec),"TriggerNames");
+  iEvent.put(std::move(hltEleObj),"HLTElectronObjects");
```
kpedro88: put behind bool

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -495,6 +495,7 @@ def makeTreeFromMiniAOD(self,process):
     )
     self.VectorInt.extend(['TriggerProducer:TriggerPass','TriggerProducer:TriggerPrescales'])
     self.VectorString.extend(['TriggerProducer:TriggerNames'])
+    self.VectorTLorentzVector.extend(['TriggerProducer:HLTElectronObjects'])
```
kpedro88: condition on `if "SingleElectron" in process.source.fileNames[0]:`

kpedro88 (changes requested):

Utils/src/TriggerProducer.cc
kpedro88: actually, just do: `(const auto& pathName : pathNamesAll)`

Utils/src/TriggerProducer.cc
kpedro88: actually, seems like this variable isn't used, so just delete it

Utils/src/TriggerProducer.cc
vhegde91: did you mean something like this?
bool saveHLTObj = true or false;
if(saveHLTObj) produces<std::vector<TLorentzVector> >("HLTElectronObjects");

this bool will be used in the next part of the code as well.

kpedro88: yes, and it should be set by the python config for the module.

vhegde91:
updated the code with most of your suggestions.
sorry, tried to make it make configurable from python, but could not

kpedro88:
@vhegde91 thanks
it should be easy to make it configurable: just do
```
saveHLTObj = iConfig.getParameter<bool>("saveHLTObj");
```
then in https://github.com/TreeMaker/TreeMaker/blob/Run2/Utils/python/triggerproducer_cfi.py, add:
```
saveHLTObj = cms.bool(False),
```
and in `makeTreeFromMiniAOD`, in `process.TriggerProducer`:
```
    saveHLTObj = cms.bool("SingleElectron" in process.source.fileNames[0])
```

Utils/src/TriggerProducer.cc
```diff
@@ -150,9 +159,29 @@ TriggerProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetu
     }
   }
 
+  edm::Handle<pat::TriggerObjectStandAloneCollection> triggerObjects;
```
kpedro88: put this entire section inside `if(saveHLTObj)`
(from here till end of for loop)

vhegde91:
@kpedro88 thanks a lot for helping with this. let me know if anything else is left.

kpedro88 (approved):

PR #269, opened 2018-02-06
Reduce CPU usage, minor tweaks

kpedro88:
* Major component of this PR: update installation recipe to reduce CPU usage by avoiding improper use of `mayConsume` - tries to prefetch nonexistent, unwanted products for every event
* Also update installation recipe to remove unneeded tags in configs, to clean up debug output
* Add option to run tracer
* Fix a few minor bugs


kpedro88:
NB, removing improper use of `mayConsume` results in a 24% (!!!) CPU reduction.

PR #270, opened 2018-02-09
Add more samples

nstrobbe:
A bunch of new samples added. 
Two existing samples updated: 
- TTTT: switch to sample with different tune since it has 10x more events
- TTZToLLNuNu_M-10: include 2 extra extensions


kpedro88:
get_py.py isn't prepared for extension samples named "\_backup\_" rather than the standard convention (why does the production team do random things like this?), so right now the file TT_TuneCUETP8M2T4_13TeV-powheg-pythia8_cff.py included in this PR is actually the file list for the "backup" sample. In this case you'll have to rerun get_py.py (just with the `--py` flag) separately for each sample and rename the "backup" file list to something appropriate so they don't clash/overwrite.

Also, can you update dict_tth, dict_diboson, and dict_singlet in https://github.com/TreeMaker/TreeMaker/blob/Run2/Production/test/condorSub?

nstrobbe:
Filelists for TT fixed, and dict_tth, dict_diboson, dict_singlet, dict_ttbar updated. 
Let me know if there is anything else missing.

kpedro88:
Do you also want me to run your private signal samples in central ntuple production?

nstrobbe:
Ah yes, that would probably be practical. 
Do I also just produce some filelists for that, and add a dict for it?

EDIT: looking at how our MINIAOD is stored, it seems like we have O(1000) files per sample. Should I merge those first?

kpedro88:
Yes, the same basic procedure can be followed. Put the file lists in the PrivateSamples folder. You can also add them to getWeightProducer if you want to have the cross-section weights.

PR #273, opened 2018-03-06
update Photon ID and EA to Summer17 v1

vhegde91:


kpedro88:
Please submit this PR to the branch Run2_2017, which will be used for 2017 production.

In addition, we need to continue to support the 2016 ID criteria for combined analyses, so please move all the parameters into Python. At some point, we will add a switch in the code to determine if a sample is 2016 or 2017.

vhegde91:
For now, should I just make PR to branch Run2_2017? Or do you want me to make some changes to python as well?
thanks!

kpedro88:
We don't have 2017 samples in the Run2_2017 branch, just code that will run in 94X.

I would prefer for all update PRs to include Python flexibility for cut values so we don't miss something and cause an inconsistency.

Also, we could probably make `effAreas` a member function of the class and just initialize it once in the constructor, instead of doing it every event.

vhegde91:
There are too many numbers to use and may be it is difficult to handle all these values from python file.
Is the following proposal OK?
- Declare different variables for these parametrs(H/E, QIso, eff area etc) in .cc file
- Use a string in the code to decide scenario and accordingly initialize these parameters. It is a kind of switch you had mentioned.
- This string is going to be configured from python file.

kpedro88:
Nope, they're all going in the python file. You can declare a vector of values as a python parameter, so not that difficult to move them all. We already do this e.g. for JetID in the GoodJetProducer.

vhegde91:
I will keep you posted as we go along.
Effective areas have been moved:
https://github.com/vhegde91/TreeMaker/blob/Run2/Utils/src/PhotonIDisoProducer.cc#L80-L139
PR will be made on Run2_2017 once I get all the other things ready.
Let me know if any changes are needed for this.

vhegde91:
Hi Kevin,
Please refer to these changes for new Photon ID:
https://github.com/vhegde91/TreeMaker/commit/e7206b2bc3aa27774260bea246a5cc19328169e6
And also this link for moving effective area eta ranges into config:
https://github.com/vhegde91/TreeMaker/commit/e9dc3e3de793250980b29bebfef9b07c78bc97f7

New code is running fine. I am going to make more checks and then make PR on Run2_2017.
Please let me know if this is fine.

kpedro88:
Hi Vinay,
This looks pretty good. A few comments:
* delete commented-out code and cout statements
* add a safety check for iso parameters that are vectors (to make sure they have the expected size)

We will need to integrate the use of Eras to set the parameters for 2016 vs. 2017. attn: @aperloff

aperloff:
@vhegde91 I'm just confirming that you're going to be making a PR to update the photon ID on the Run2_2017 branch. I was going to do that, but I'm happy to work on other updates if you are already working on this.

vhegde91:
Updated PR is here: https://github.com/TreeMaker/TreeMaker/pull/287
I guess code taken from branch Run2_2017 does not work on 2016 MC?

PR #274, opened 2018-03-22
Add two prescaled triggers for efficiency measurement

nstrobbe:


kpedro88:
Included in #281 for 2017 branch, I don't think we'll make another set of ntuples with the 2016/80X branch.

PR #282, opened 2018-04-03
TreeMaker Eras to handle settings for different years

kpedro88:


aperloff:
I don't see a problem with this. My only question would be why have an "era" and a "localera"?

kpedro88:
The "era" is the CMSSW one, which may be needed to configure certain modules that get pulled in by jet/met algorithms. The "localera" is our internal one. (We could try to reuse CMSSW modifiers, but it seems easier just to make our own.)

PR #284, opened 2018-04-10
MET and ZinvBkg

aperloff:
Add new 2017 MET filters. Also fix the soft drop mass and subjet collections for the ZinvBkg configuration.

### DIFF
--- a/TreeMaker/python/doZinvBkg.py
+++ b/TreeMaker/python/doZinvBkg.py
@@ -46,9 +46,11 @@ def reclusterZinv(self, process, cleanedCandidates, suff):
 
     # update some userfloat names
     process.JetPropertiesAK8Clean.prunedMass = cms.vstring('ak8PFJetsCHSCleanPrunedMass')
+    process.JetPropertiesAK8Clean.softDropMass = cms.vstring('SoftDrop')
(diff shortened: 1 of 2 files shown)


aperloff:
I didn't update all of ZinvBkg to use PUPPI jets as they were using "cleaned" jets. Is your wish to completely change that? Should I remove the pruned collections and have it simply use the preexisting ak8puppi jets?

kpedro88:
The Zinv bkg should use the cleaned version of whatever jets are used in the central collection - now this is puppi for AK8, so we need to make a cleaned AK8 puppi collection.

aperloff:
Okay, can you merge this PR in and then I'll do a separate PR to change all jet collections to PUPPI? That will just help me a little since I have several atomic branches right now, but a change that big is going to need to merge in changes from several of those.

kpedro88:
all *AK8* jet collections, you can leave the AK4 ones alone

aperloff:
@kpedro88 So you're saying leave the AK4 collection as "slimmedJets" (made from ak4PFJetsCHS) and the AK8 collection as "slimmedJetsAK8" (made from ak8PFJetsPUPPI). If I'm not changing the AK4 jet collection then I guess, though I will confirm, that the only changes will need to be to the AK8Cleaned collection (the one we recluster using the JetToolbox).

kpedro88:
Yes, I think that's correct.

PR #285, opened 2018-04-11
Update jet criteria

aperloff:
Update JetID
Update makeGoodJets to be cone size aware (i.e. minimum pT for AK8 jets is now 170).

PR #287, opened 2018-04-12
Updating photon ID and EA for 2017 on Run2_2017

vhegde91:
Updated to 2017 data photon IDs and new effective area. Make the ID, ISO and EA values to be configurable from python file.


kpedro88:
This branch seems broken somehow. There are commits and reversions that shouldn't be needed. Can you clean up the branch so it only has clean commits with the final desired changes?

Also, we need to use the Era functionality to retain support for the 2016 ID for appropriate scenarios. @aperloff, I can merge this and then you can make a followup PR with the Eras, or you can work with Vinay privately to include it in this PR. Let me know.

Utils/src/PhotonIDisoProducer.cc
```diff
@@ -100,6 +107,22 @@ PhotonIDisoProducer::PhotonIDisoProducer(const edm::ParameterSet& iConfig):
   debug(iConfig.getUntrackedParameter<bool>("debug",true))
 {
 
+  effArEtaLow_           = iConfig.getParameter <std::vector<double>> ("effArEtaLow");
```
kpedro88, line 110: all of these can be initialized in the constructor initialization list (like the input tags already are)

Utils/src/PhotonIDisoProducer.cc
```diff
@@ -121,6 +144,29 @@ PhotonIDisoProducer::PhotonIDisoProducer(const edm::ParameterSet& iConfig):
   produces< std::vector< bool > >("fullID");
   produces< std::vector< bool > >("electronFakes");
   produces< bool >("hasGenPromptPhoton");
+  if(effArEtaLow_.size()!=7 || effArEtaHigh_.size()!=7
```
kpedro88, line 147: this can just be:
`bool correctSizeVecs = !(effArEtaLow_.size()!=7 || effArEtaHigh_.size()!=7 || effArChHad_.size()!=7 || effArNuHad_.size()!=7 || effArGamma_.size()!=7 || pfNuIsoRhoCorr_EB_cut_.size()!=3 || pfNuIsoRhoCorr_EE_cut_.size()!=3 || pfGmIsoRhoCorr_EB_cut_.size()!=2 || pfGmIsoRhoCorr_EE_cut_.size()!=2);`

Utils/src/PhotonIDisoProducer.cc
```diff
@@ -121,6 +144,29 @@ PhotonIDisoProducer::PhotonIDisoProducer(const edm::ParameterSet& iConfig):
...
+     || effArChHad_.size()!=7 || effArNuHad_.size()!=7 || effArGamma_.size()!=7 
+     || pfNuIsoRhoCorr_EB_cut_.size()!=3 || pfNuIsoRhoCorr_EE_cut_.size()!=3 
+     || pfGmIsoRhoCorr_EB_cut_.size()!=2 || pfGmIsoRhoCorr_EE_cut_.size()!=2)  correctSizeVecs = false;
+  else correctSizeVecs = true;
+  if(debug || !correctSizeVecs){
```
kpedro88, line 152: if the vector sizes are not correct, you should throw an exception so the code stops. Do not try to run with bad parameter input.

Utils/src/PhotonIDisoProducer.cc
```diff
@@ -239,39 +269,44 @@ PhotonIDisoProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::Event
...
     }
-    
-    // apply id cuts using https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedPhotonIdentificationRun2#Recommended_Working_points_for_2
-    if (isBarrelPhoton) {
-      if (iPhoton->hadTowOverEm() < 0.0597 && !hasMatchedPromptElectron(iPhoton->superCluster(), electrons, conversions, beamSpot->position())) {
-	passIDLoose = true;
-	if (sieie < 0.01031) {
-	  passID = true;
+
+    if(correctSizeVecs){
```
kpedro88, line 273: should never get to this point with incorrect vector sizes (`correctSizeVecs` doesn't even need to be a member variable if above recommendation is followed)

aperloff:
@vhegde91 @kpedro88 Assuming the C++ changes aren't too substantial, I think it might be easier to make a followup PR with eras. Alternatively if Vinay wants to try to implement them he can take a look at #285 to see how I included them.

vhegde91:
@aperloff if it is possible and easier for you to take it from here, then that would be very useful. Sorry, I don't think I can implement them, without making it even more messy. I am not sure how I would clean up the old and unnecessary commits.

kpedro88:
@vhegde91 the interest of time, I'll remake your branch with the commits squashed. But in the future, you should follow https://twiki.cern.ch/twiki/bin/viewauth/CMS/CMSGitTutorial to learn more about using git to manage history.

kpedro88:
@vhegde91 after looking closer, this branch is too messed up to be worth fixing. Just copy your versions of the files TreeMaker/python/doZinvBkg.py and  Utils/src/PhotonIDisoProducer.cc and commit them in a new area/branch.

vhegde91:
sorry for this mess. Should I make new changes that you suggested in the code and then commit or should I just commit the same code whatever is shown here on a different branch?

kpedro88:
I would commit what is already here and then make the updates.

vhegde91:
would you like to close this PR and then get a clean PR from here? https://github.com/TreeMaker/TreeMaker/compare/Run2_2017...vhegde91:new2

kpedro88:
Yes, please do that

PR #288, opened 2018-04-12
adding 2017 photon ID

vhegde91:


Utils/src/PhotonIDisoProducer.cc
```diff
@@ -100,6 +107,22 @@ PhotonIDisoProducer::PhotonIDisoProducer(const edm::ParameterSet& iConfig):
   debug(iConfig.getUntrackedParameter<bool>("debug",true))
 {
 
+  effArEtaLow_           = iConfig.getParameter <std::vector<double>> ("effArEtaLow");
```
kpedro88: all of these can be initialized in the constructor initialization list (like the input tags already are)

Utils/src/PhotonIDisoProducer.cc
```diff
@@ -121,6 +144,29 @@ PhotonIDisoProducer::PhotonIDisoProducer(const edm::ParameterSet& iConfig):
   produces< std::vector< bool > >("fullID");
   produces< std::vector< bool > >("electronFakes");
   produces< bool >("hasGenPromptPhoton");
+  if(effArEtaLow_.size()!=7 || effArEtaHigh_.size()!=7
```
kpedro88: this can just be:
`bool correctSizeVecs = !(effArEtaLow_.size()!=7 || effArEtaHigh_.size()!=7 || effArChHad_.size()!=7 || effArNuHad_.size()!=7 || effArGamma_.size()!=7 || pfNuIsoRhoCorr_EB_cut_.size()!=3 || pfNuIsoRhoCorr_EE_cut_.size()!=3 || pfGmIsoRhoCorr_EB_cut_.size()!=2 || pfGmIsoRhoCorr_EE_cut_.size()!=2);`

Utils/src/PhotonIDisoProducer.cc
```diff
@@ -121,6 +144,29 @@ PhotonIDisoProducer::PhotonIDisoProducer(const edm::ParameterSet& iConfig):
...
+     || effArChHad_.size()!=7 || effArNuHad_.size()!=7 || effArGamma_.size()!=7 
+     || pfNuIsoRhoCorr_EB_cut_.size()!=3 || pfNuIsoRhoCorr_EE_cut_.size()!=3 
+     || pfGmIsoRhoCorr_EB_cut_.size()!=2 || pfGmIsoRhoCorr_EE_cut_.size()!=2)  correctSizeVecs = false;
+  else correctSizeVecs = true;
+  if(debug || !correctSizeVecs){
```
kpedro88: if the vector sizes are not correct, you should throw an exception so the code stops. Do not try to run with bad parameter input.

Utils/src/PhotonIDisoProducer.cc
```diff
@@ -239,39 +269,44 @@ PhotonIDisoProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::Event
...
     }
-    
-    // apply id cuts using https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedPhotonIdentificationRun2#Recommended_Working_points_for_2
-    if (isBarrelPhoton) {
-      if (iPhoton->hadTowOverEm() < 0.0597 && !hasMatchedPromptElectron(iPhoton->superCluster(), electrons, conversions, beamSpot->position())) {
-	passIDLoose = true;
-	if (sieie < 0.01031) {
-	  passID = true;
+
+    if(correctSizeVecs){
```
kpedro88: should never get to this point with incorrect vector sizes (`correctSizeVecs` doesn't even need to be a member variable if above recommendation is followed)

Utils/src/PhotonIDisoProducer.cc
```diff
@@ -77,28 +77,49 @@ class PhotonIDisoProducer : public edm::global::EDProducer<> {
...
-  conversionTok_(consumes<std::vector<reco::Conversion>>(conversionCollection)),
-  beamspotTok_(consumes<reco::BeamSpot>(beamspotCollection)),
-  ecalRecHitsInputTag_EE_Token_(consumes<EcalRecHitCollection>(ecalRecHitsInputTag_EE_)),
-  ecalRecHitsInputTag_EB_Token_(consumes<EcalRecHitCollection>(ecalRecHitsInputTag_EB_)),
-  rhoTok_(consumes<double>(rhoCollection)),
-  genParTok_(consumes<edm::View<reco::GenParticle>>(genParCollection)),
-  debug(iConfig.getUntrackedParameter<bool>("debug",true))
-{
+PhotonIDisoProducer::PhotonIDisoProducer(const edm::ParameterSet& iConfig){
+  photonCollection       = iConfig.getUntrackedParameter<edm::InputTag>("photonCollection");
```
kpedro88: this is the exact opposite of what I wanted you to do

vhegde91: I looked at what Alex was pointing to and made changes according to that : https://github.com/aperloff/TreeMaker/blob/2fc19aff65d33facd109443b5fd10d9bb3f67220/Utils/src/GoodJetsProducer.cc#L85-L118

kpedro88: well, it should be improved there too. Please undo this change and put all your new parameters into the initialization list (this comes *before* the `{` in the constructor).

vhegde91: I could implement for vector<double>. But for double it gives the error when I tried this:
`debug(iConfig.getUntrackedParameter<bool>("debug",true)),
  hadTowOverEm_EB_cut_(iConfig.getParameter<double>("hadTowOverEm_EB_cut")),
`
While compiling it gives this error: "/uscms_data/d3/vhegde/Physics/makeRA2Trees_V13/for2017/CMSSW_8_0_30/src/TreeMaker/Utils/src/PhotonIDisoProducer.cc: In constructor 'PhotonIDisoProducer::PhotonIDisoProducer(const edm::ParameterSet&)':
/uscms_data/d3/vhegde/Physics/makeRA2Trees_V13/for2017/CMSSW_8_0_30/src/TreeMaker/Utils/src/PhotonIDisoProducer.cc:84:10: error: 'PhotonIDisoProducer::hadTowOverEm_EE_cut_' will be initialized after [-Werror=reorder]
   double hadTowOverEm_EE_cut_, sieie_EE_cut_, pfChIsoRhoCorr_EE_cut_;"
........
Any idea why it is so?

kpedro88: Exactly what the error says: you have to initialize in the order the member variables are declared in the class.

vhegde91: I never expected that the order matters.
This is how the final code looks like: 
`pfGmIsoRhoCorr_EB_cut_(iConfig.getParameter <std::vector<double>> ("pfGmIsoRhoCorr_EB_cut")),
 pfGmIsoRhoCorr_EE_cut_(iConfig.getParameter <std::vector<double>> ("pfGmIsoRhoCorr_EE_cut"))
{`
Is there any other changes that are needed? Should I just commit these changes? If so, any specific message for the commit?

kpedro88: If you have made the change to put everything that was in the initializer list back in the initializer list, and to add all of the new parameters (not just those two you listed), please commit it.

vhegde91:
@kpedro88 Thanks for all the help!

PR #291, opened 2018-04-18
Minor changes to the HadTauBkg code

aperloff:
Update formatting, btag discriminators, and redundant configuration parameters.

I couldn't find any configuration parameters for tau ID, if they even exist. There was some vague reference to tau ID being updated, but I'm not sure it's important here. If I've missed something please let me know and I can add it to the PR.


kpedro88:
We don't evaluate or store the actual tau ID in TreeMaker. It was considered once, but found not to be significantly more efficient as a veto than isolated pion tracks.

PR #295, opened 2018-04-25
Optimize TriggerProducer

kpedro88:
`TriggerProducer` was in the top 5 in a CPU test with the 2017 scenario. Hash tables make O(N^2) into O(N)! For the current list of triggers (containing both 2016 and 2017 lines), this is a ~10x speedup.

Also, we had been storing the version number in the TriggerName string if the trigger was found in the event. Instead, we can store the version number in a separate branch, so the TriggerName branch is always the same (in case of future optimizations for storing metadata).

Fixed some bugs in the trigger name list revealed by these changes and added a comment to clarify the details.


kpedro88:
Another optimization: make the map from the input list of triggers, and just compare all the per-event triggers to it. Should result in fewer string hashes (and allocations) overall, and seems to be at least ~50% faster than the previous way.

PR #298, opened 2018-04-27
Code checks

kpedro88:
CMSSW can run code checks via clang-tidy in scram. The default list is here:
https://github.com/cms-sw/cmssw/blob/master/.clang-tidy

I added a few more: `cppcoreguidelines-pro-type-member-init, modernize-loop-convert`

and then ran over TreeMaker source code as follows:
```
touch TreeMakerCode.txt
find TreeMaker/ -name "*.h" >> TreeMakerCode.txt
find TreeMaker/ -name "*.cc" >> TreeMakerCode.txt
env USER_CODE_CHECKS_FILE=TreeMakerCode.txt scram b code-checks
```

The code is automatically updated to conform to the specified code checks.

A few caveats about `modernize-loop-convert`:
* doesn't always use `const` where it could
* sometimes chooses unclear names for loop variables
* doesn't seem to recognize range-based loops for `edm::View` types

I dealt with these few issues by hand.

### DIFF
--- a/TreeMaker/interface/TreeMaker.h
+++ b/TreeMaker/interface/TreeMaker.h
@@ -46,13 +46,13 @@ class TreeObjectBase;
 class TreeMaker : public edm::one::EDAnalyzer<edm::one::SharedResources> {
 	public:
 		explicit TreeMaker(const edm::ParameterSet&);
-		~TreeMaker();
+		~TreeMaker() override;
 		static void fillDescriptions(edm::ConfigurationDescriptions& descriptions);
 
 	private:
-		virtual void beginJob() override;
-		virtual void analyze(const edm::Event&, const edm::EventSetup&) override;
-		virtual void endJob() override;
+		void beginJob() override;
+		void analyze(const edm::Event&, const edm::EventSetup&) override;
+		void endJob() override;
 		// ----------member data ---------------------------
 		edm::Service<TFileService> fs;
 		string treeName;
@@ -62,9 +62,9 @@ class TreeMaker : public edm::one::EDAnalyzer<edm::one::SharedResources> {
 		vector<TreeTypes> VarTypes;
 		map<string,unsigned> nameCache;
 		// general event information
-		UInt_t runNum;
-		UInt_t lumiBlockNum;
-		ULong64_t evtNum;
+		UInt_t runNum{};
+		UInt_t lumiBlockNum{};
+		ULong64_t evtNum{};
 		vector<TreeObjectBase*> variables;
 };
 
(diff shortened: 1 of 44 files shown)


TreeMaker/interface/TreeMaker.h
```diff
@@ -62,9 +62,9 @@ class TreeMaker : public edm::one::EDAnalyzer<edm::one::SharedResources> {
 		vector<TreeTypes> VarTypes;
 		map<string,unsigned> nameCache;
 		// general event information
-		UInt_t runNum;
-		UInt_t lumiBlockNum;
-		ULong64_t evtNum;
+		UInt_t runNum{};
```
aperloff, line 65: Why add brackets here? What does that do in a technical sense?

Utils/src/METDouble.cc
```diff
@@ -142,7 +142,7 @@ METDouble::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetup& iSe
    double dpnhat[3];
    unsigned int goodcount=0;
    
-   for(int i=0; i<3; ++i)dpnhat[i]=-999;
+   for(double & i : dpnhat)i=-999;
```
aperloff, line 145: Why not 'auto' here?

Utils/src/METDouble.cc
```diff
@@ -220,8 +220,8 @@ METDouble::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetup& iSe
    auto htp5 = std::make_unique<double>(dpnhat[2]);
    iEvent.put(std::move(htp5),"DeltaPhiN3");
    float mindpn=9999;
-   for(int i=0; i<3; ++i){
-      if(mindpn>fabs(dpnhat[i]))mindpn=fabs(dpnhat[i]);
+   for(double i : dpnhat){
```
aperloff, line 223: Same question.

TreeMaker/interface/TreeMaker.h
kpedro88: see:
* https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-pro-type-member-init.html
* http://en.cppreference.com/w/cpp/language/zero_initialization

Utils/src/METDouble.cc
kpedro88: auto is often not used for primitive types

aperloff: Really. I didn't know that. That's good information to have.

aperloff:
Looks good to me.

Utils/src/METDouble.cc
kpedro88: The goal of auto is to improve readability. Obviously this is somewhat in the eye of the beholder, but while replacing some 50-character iterator type with auto definitely improves readability, replacing int/double/float etc. probably doesn't.

PR #299, opened 2018-04-30
Update 2016/2017 JER and 2016 JEC

aperloff:
Update the JER used for both the 2016 and 2017 MC. Also update the JEC used for the ReMiniAOD of the 2016 data and MC. Three new scenarios were created for the ReMiniAOD of the 2016 samples. The ReMiniAOD processing hasn't started yet and no 94X global tag has been defined. The GT will need to be updated later.


Production/python/scenarios.py
```diff
@@ -80,13 +80,46 @@ def __init__(self,sname):
...
+                globaltag="80X_mcRun2_asymptotic_2016_TrancheIV_v6",
+                tagname="PAT",
+                geninfo=True,
+                jecfile="data/jec/Summer16_07Aug2017_V9_MC",
+                jerfile="data/jer/Summer16_25nsV1_MC",
+                pufile="TreeMaker/Production/test/data/PileupHistograms_0721_63mb_pm5.root",
+                era="Run2_2016",
+                localera="TM2016",
+            )
+        elif sname == "Summer16sigLegacyReMiniAOD":
```
kpedro88: I would put "sig" at the end

PR #301, opened 2018-05-06
New approach for storing jet systematic information

kpedro88:
Brief ntuple size (kB/event) history:

| Type | V12 | V14 | V15a | V15b |
|--------|------:|-----:|-----:|-----:|
| MC (ttbar) | 6.3 | 5.5 | 9.0 | 5.5 |
| Data (MET) | 1.6 | 1.0 | ? | ? |

From V12 to V14, we stopped storing the soft jets (pT < 30 GeV).

For the upcoming V15, keeping 5 copies of all of the AK4 jet properties (vector branches) ends up taking a significant amount of space (see V15a). (Previously, we only kept a few property branches, but now more are needed e.g. for top tagging or b-tagging studies.)

Instead, we can store the list of original indices for the systematically varied jet collections (which get resorted by pT), which removes the need to duplicate the property branches. For a further optimization, we don't even need to store the 4-vectors for the systematic collections (saving 0.5 kB/evt), because we already store the JEC and JER uncertainty factors (which can easily be used to rescale the central 4-vectors). This way, the most novel operation (sorting the jet collection) doesn't have to be repeated, but no information is duplicated. (We still keep the scalar computed quantities, like NJets etc., which also involve some potentially nontrivial computations.)

However, to make this work, we once again need to store all the jets down to the miniAOD cut of pT > 10 GeV. This causes the size of the central jets collection to increase substantially, but there's still an overall saving by eliminating the systematic collections (see V15b). It also enables studies with lower-pT jets, which may be desired by some analyses. (I tried just going down to pT > 20 GeV, but there are still some jets below 20 GeV that can fluctuate up above 30 GeV with the systematics.)

Another consideration is data vs MC. In V12 we have approximately 2x as many data events compared to MC. We don't store systematic variations for data, so including the soft jets down to 10 GeV will just increase the size. But even if we assume the data event size goes back to 1.6 kB, the MC size decreases enough that we save space overall. It is probably worthwhile to have consistent pT cuts in data and MC to avoid confusion. We should double-check this once we have 94X miniAODv2 data files available.

### DIFF
--- a/TreeMaker/python/makeJetVars.py
+++ b/TreeMaker/python/makeJetVars.py
@@ -1,14 +1,14 @@
 import FWCore.ParameterSet.Config as cms
 from TreeMaker.TreeMaker.addJetInfo import addJetInfo
 
-def makeGoodJets(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInputTag(), jetConeSize=0.4):
+def makeGoodJets(self, process, JetTag, suff, storeProperties, systematic=False, SkipTag=cms.VInputTag(), jetConeSize=0.4):
     from TreeMaker.TreeMaker.TMEras import TMeras
     from TreeMaker.Utils.goodjetsproducer_cfi import GoodJetsProducer
-    GoodJets = GoodJetsProducer.clone()
-    TMeras.TM2016.toModify(GoodJets,
+    GoodJets = GoodJetsProducer.clone(
         TagMode                   = cms.bool(True),
         JetTag                    = JetTag,
-        jetPtFilter               = cms.double(20 if self.fastsim else 30),
+        # keep lower-pt central jets in case they fluctuate up in systematic collections
+        jetPtFilter               = cms.double(170 if jetConeSize==0.8 else 30 if systematic else 0),
         ExcludeLepIsoTrackPhotons = cms.bool(True),
         JetConeSize               = cms.double(jetConeSize),
         SkipTag                   = SkipTag,
@@ -16,8 +16,6 @@ def makeGoodJets(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInpu
         SaveAllJetsPt             = False, # exclude low pt jets from good collection
     )
     TMeras.TM2017.toModify(GoodJets,
-        TagMode                   = cms.bool(True),
-        JetTag                    = JetTag,
         maxNeutralFraction        = cms.double(0.90),
         maxNeutralFractionHE      = cms.double(1.00), #Turned off as not needed for the tight WP
         minNeutralFractionHF      = cms.double(0.02),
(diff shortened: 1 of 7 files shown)


TreeMaker/python/makeJetVars.py
```diff
@@ -4,33 +4,25 @@
 def makeGoodJets(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInputTag(), jetConeSize=0.4):
     from TreeMaker.TreeMaker.TMEras import TMeras
     from TreeMaker.Utils.goodjetsproducer_cfi import GoodJetsProducer
-    GoodJets = GoodJetsProducer.clone()
-    TMeras.TM2016.toModify(GoodJets,
+    GoodJets = GoodJetsProducer.clone(
         TagMode                   = cms.bool(True),
         JetTag                    = JetTag,
-        jetPtFilter               = cms.double(20 if self.fastsim else 30),
+        jetPtFilter               = cms.double(170 if jetConeSize==0.8 else 30),
```
aperloff, line 10: What happened to all of the fastsim parts? Does this not matter because you'll need all jet pt even for fastsim?

Utils/src/JetUncertaintyProducer.cc
```diff
@@ -71,7 +71,8 @@ void JetUncertaintyProducer::produce(edm::StreamID, edm::Event& iEvent, const ed
 		edm::Ptr<pat::Jet> jetPtr = jets->ptrAt(idx);
 		pat::Jet ajet(jetPtr);
 		math::XYZTLorentzVector vjet = ajet.p4();
-		
+		ajet.addUserInt("origIndex",idx);
```
aperloff, line 72: I'm really commenting on line 69. How does computing the idx every time, but using and iterator for the "jets" handle save any time. Isn't std::distance slow. Wouldn't it be better to simply compute the size of the jets vector once and use the for loop index? This might be inconsequential because we don't expect that many jets, but I felt like mentioning it.

aperloff (commented):
So this means there will be the main jets collection, with properties and the origIndex int, and the systematic collections without any properties, but with the origIndex? Do I have that right?

aperloff:
I like the space saving, but I'm confused about what is being sacrificed. So will it be up to analyzers to scale jets by JER/JEC, or is the scaled pT the one thing being saved now?

TreeMaker/python/makeJetVars.py
kpedro88: Yes, in fact especially for fastsim, since that's one of the places we always evaluate JEC/JER uncertainty.

Utils/src/JetUncertaintyProducer.cc
kpedro88: I think I copied this from somewhere in CMSSW. Might be worth making the minor optimization you suggest.

kpedro88:
The origIndex branches will only be added for the systematic collections, with nothing else. The value of each entry in the origIndex branch will point to the central collections (essentially just keeping track of the sorting).

Yes, the analyzers will have to scale the 4-vectors by the uncertainty factors (which are already stored as branches in the central collection). Originally I was considering keeping the systematic 4-vector branches, but then we end up saving 5 copies of the jet collection down to pT > 10 (or if we put a pT cut on the systematic collections, the utility is limited for future studies). In practice I think a lot of people were using the factors anyway, or the skims in which I've already applied the variations.

aperloff:
Sounds good to me. I like it.

Utils/src/JetUncertaintyProducer.cc
kpedro88: Tested a change to this and it's not any faster. I'll commit it anyway just to simplify the code.

aperloff: Both interesting and good to know. Thanks for looking into it.

PR #308, opened 2018-05-22
add puppi speedup branch to setup

kpedro88:
see https://github.com/cms-sw/cmssw/pull/23270 for details

In the future, we should study if reusing the default miniAOD puppi weights is sufficient for the Zinv background estimation in the H(bb)+MET analysis (even though a photon or dilepton is removed from the PF candidate collection).


kpedro88:
posted #309 to track the study of reusing puppi weights

PR #313, opened 2018-06-20
Adding deepAK8 to the nTuples

cmadrid1:


Utils/src/DeepAK8Producer.cc
```diff
@@ -0,0 +1,73 @@
...
+    fatjetNN_(nullptr)
+{
+    // Initialize the FatJetNN class in the constructor
+    auto cc = consumesCollector();
+    fatjetNN_ = new deepntuples::FatJetNN(iConfig, cc);
+    // Load json for input variable transformation
+    fatjetNN_->load_json("preprocessing.json"); // use the full path or put the file in the current working directory (i.e., where you run cmsRun)
+    // Load DNN model and parameter files
+    fatjetNN_->load_model("resnet-symbol.json", "resnet.params"); // use the full path or put the file in the current working directory (i.e., where you run cmsRun)
+}
```
kpedro88: the producer needs to declare what it produces in the constructor

Utils/src/DeepAK8Producer.cc
```diff
@@ -0,0 +1,73 @@
...
+void DeepAK8Producer::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSetup& iSetup) const
+{
+    // Need to access event info
+    fatjetNN_->readEvent(iEvent, iSetup);
+
+    // Get, loop over the jets, and fill  
+    edm::Handle<edm::View<pat::Jet>> jetDeepAK8;
+    iEvent.getByToken(JetAK8Tok_, jetDeepAK8);
+
+    auto tDiscriminatorDeep = std::make_unique<std::vector<double>>();
```
kpedro88: It would be better to produce these as userfloats. For an example, see https://github.com/TreeMaker/TreeMaker/blob/Run2_2017/Utils/src/JetUncertaintyProducer.cc (follow the `produces<edm::ValueMap<float>>("");` part).

Then, you can add this module to the python config after the double b-tagger, and update the jet collection to know about the new userfloats by modifying the `addJetInfo` call:
https://github.com/TreeMaker/TreeMaker/blob/0a9d941d505822f5537a0b6fa67a3e5fa07d15ad/TreeMaker/python/makeTreeFromMiniAOD_cff.py#L731

setup.sh
```diff
@@ -19,6 +19,12 @@ scram project ${CMSSWVER}
 cd ${CMSSWVER}/src/
 # cmsenv
 eval `scramv1 runtime -sh`
+git clone https://gitlab.cern.ch/DeepAK8/NNKit.git
+cp NNKit/misc/mxnet_predict.xml $CMSSW_BASE/config/toolbox/$SCRAM_ARCH/tools/selected
+scram setup mxnet_predict
+rm $CMSSW_BASE/external/$SCRAM_ARCH/lib/libmxnet_predict.so
```
kpedro88: it seems like the `rm` isn't necessary, since the `cp` on the next line will automatically overwrite the existing library.

kpedro88:
@cmadrid1 to test this (on 2017 miniAODv2 relvals), you can run the command (in directory `TreeMaker/Production/test`):
```
python unitTest.py test=8 run=1
```

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -729,8 +729,11 @@ def makeTreeFromMiniAOD(self,process):
     process.load("RecoBTag.SecondaryVertex.candidateBoostedDoubleSecondaryVertexAK8Computer_cfi")
     process.load("RecoBTag.SecondaryVertex.pfBoostedDoubleSecondaryVertexAK8BJetTags_cfi")
 
+    from TreeMaker.Utils.deepak8producer_cfi import DeepAK8Producer
+    process.deepAK8 = DeepAK8Producer.clone()
+
     # add discriminator and update tag
-    process, JetAK8Tag = addJetInfo(process, JetAK8Tag, [], [], cms.VInputTag(cms.InputTag("pfBoostedDoubleSecondaryVertexAK8BJetTags")))
+    process, JetAK8Tag = addJetInfo(process, JetAK8Tag, [], [], cms.VInputTag(cms.InputTag("pfBoostedDoubleSecondaryVertexAK8BJetTags")), ['deepAK8'])
```
kpedro88: should be:
```
process, JetAK8Tag = addJetInfo(process, JetAK8Tag, ['deepAK8:tDiscriminatorDeep','deepAK8:wDiscriminatorDeep'], [], cms.VInputTag(cms.InputTag("pfBoostedDoubleSecondaryVertexAK8BJetTags")))
```
but this will only work once the producer outputs userfloats instead of `vector<double>`

TreeMaker/python/makeJetVars.py
```diff
@@ -328,6 +330,8 @@ def makeJetVarsAK8(self, process, JetTag, suff, storeProperties):
         JetPropertiesAK8.bDiscriminatorCSV = cms.vstring('pfBoostedDoubleSecondaryVertexAK8BJetTags')
         JetPropertiesAK8.subjets = cms.vstring('SoftDropPuppi')
         JetPropertiesAK8.SJbDiscriminatorCSV = cms.vstring('SoftDropPuppi','pfCombinedInclusiveSecondaryVertexV2BJetTags')
+        JetPropertiesAK8.tDiscriminatorDeep = cms.vstring('tDiscriminatorDeep')
```
kpedro88: The vstring value should be `'deepAK8:tDiscriminatorDeep'` (and similar for the next one)

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -729,8 +729,11 @@ def makeTreeFromMiniAOD(self,process):
     process.load("RecoBTag.SecondaryVertex.candidateBoostedDoubleSecondaryVertexAK8Computer_cfi")
     process.load("RecoBTag.SecondaryVertex.pfBoostedDoubleSecondaryVertexAK8BJetTags_cfi")
 
+    from TreeMaker.Utils.deepak8producer_cfi import DeepAK8Producer
```
kpedro88, line 732: please `git add` the file `TreeMaker/Utils/python/deepak8producer_cfi.py`

setup.sh
```diff
@@ -33,3 +39,6 @@ git clone git@github.com:${FORK}/TreeMaker.git -b ${BRANCH}
 scram b -j 8
 cd TreeMaker/Production/test/condorSub/
 python $CMSSW_BASE/src/Condor/Production/python/linkScripts.py
+ln -s ${CMSSW_BASE}/src/NNKit/data/ak8/full/preprocessing.json ${CMSSW_BASE}/src/TreeMaker/Production/test/.
```
kpedro88: it would be better to link these into `TreeMaker/Production/test/data` (where we keep other various input files)

setup.sh
```diff
@@ -19,6 +19,12 @@ scram project ${CMSSWVER}
 cd ${CMSSWVER}/src/
 # cmsenv
 eval `scramv1 runtime -sh`
+echo "Needs your CERN username and pass word: NNKit is being cloned from gitlab"
```
kpedro88: This is quite unfortunate. It means the setup script can no longer be automated. We should look for ways around this.

setup.sh
```diff
@@ -19,6 +19,12 @@ scram project ${CMSSWVER}
...
+git clone https://gitlab.cern.ch/DeepAK8/NNKit.git
+cp NNKit/misc/mxnet_predict.xml $CMSSW_BASE/config/toolbox/$SCRAM_ARCH/tools/selected
+scram setup mxnet_predict
+cp NNKit/misc/lib/libmxnet_predict.so $CMSSW_BASE/external/$SCRAM_ARCH/lib/libmxnet_predict.so
+
 git cms-init
```
kpedro88: you need to do `git cms-init` before doing anything else in the `$CMSSW_BASE/src` directory

cmadrid1:
I think this is in a "good" state. Although, I'm getting an error with a different variable when I try to test it. The code at least creates the DeepAK8Producer object before failing on something unrelated (at least I think it is unrelated).

kpedro88:
@cmadrid1 you still need to modify the producer to actually make the value maps (see comment https://github.com/TreeMaker/TreeMaker/pull/313#discussion_r196641127)

cmadrid1:
@kpedro88 Right, I forgot to do that

kpedro88:
@cmadrid1 please accept my PR https://github.com/StealthStop/TreeMaker/pull/5 to your branch, which includes a few minor fixes.

(The unrelated error you saw occurs because I updated one of our CMSSW patches and the corresponding TreeMaker changes are in #312, which isn't merged yet. I tested with that PR as well and was able to get things to work.)

kpedro88:
I also ran a CPU profile with the DeepAK8 producer included. It is unbelievably slow, increasing our total CPU usage by 50%. I am looking at https://mxnet.incubator.apache.org/faq/perf.html to see if there are feasible ways to make the inference faster, and may also contact the DeepAK8 developers.

cmadrid1:
@kpedro88 I just merged your PR and it is good to know that it is working.  Let me know if there is anything else you need me to do.  Thank you.

PR #319, opened 2018-07-02
Update the unit tests for 2017 MC and data.

aperloff:
Also update the name used for the 2017 data scenario (the rereco has changed).


kpedro88:
can you update the scenario name in the readme also?

aperloff:
I did already.

PR #332, opened 2018-07-20
Includes changes used to run the HEM RelVal Production

aperloff:
This commit includes the files and changes needed produce ntuples to study the HEM issue. This production is being called V15a and includes both MC and data relval samples. It's not the cleanest PR in terms of commits, but it isn't too bad either. Only some of these changes will be ported to the Run2_2017 branch, but it looks like those changes will need to be made by hand.

The changes to be back-ported are:
    - Remove the hard coded HLT tagname from prescaleweightproducer
    - Changes to the MCSample class
    - Changes to getWeightProducer


aperloff:
Any way to easily remove the site and weight files from within the Production/python/<production>/ folders? I just noticed that they are there.

kpedro88:
Figure out which commit accidentally included them, then rebase it to avoid adding the files.

PR #334, opened 2018-07-31
Move to CMSSW_9_4_9, MET fix, DeepAK8 updates, etc.

kpedro88:
CMSSW_9_4_9:
* JER and puppi speedups now included in base release, removed branches from setup

Bug fixes:
* reject all non-PF jets (in case new JECs move jets w/ pT < 170 above threshold)
* fix duplicated class dictionary for `vector<int>`

MET fix & other improvements:
* 2017F EE PF threshold MET fix implemented as option in MET tool, branch included in setup.sh
* The jet collection produced by the fix (excluding jets w/ pT < 75 GeV, 2.65 < |eta| < 3.139) can also be used to compute MHT
* MET fix applied by default for all 2017 MC and data, "orig" variables for MET, METPhi, MHT, MHTPhi, DeltaPhi1/2/3/4 are stored (and also for the corresponding Zinv clean variables)
* disabled JEC reapplication in the MET tool (we already do it)
* disabled reclustering when recomputing MET from candidates (we already did this previously - it was included in the 8_0_X version of the MET tool, but then lost or reverted in the 9_4_X version. I fixed it.)
* disabled unnecessary MET significance computations (not needed when we rerun the tool to get the "orig" values)
* sped up MET significance computation by ~40%

DeepAK8:
* moved to mxnet 1.2.1.mod1 w/ thread-local engine patch
* disable LPC cvmfs usage (not broadly available on the grid, library out of date)
* add decorrelated version of tagger and associated discriminators

Some notes on DeepAK8 performance:
* newest mxnet version is ~10-15% faster
* When the tagger is disabled, we get ~11 events/sec.
* When the tagger is enabled with the current configuration (run twice, once for full and once for decorrelated), we get ~6 events/sec. In other words, we spend ~40% of our CPU time per event just evaluating DeepAK8.
* According to a [presentation](https://indico.cern.ch/event/746010/contributions/3083337/attachments/1694525/2727184/deepAK8_cmssw_integration_RECOAT_20180727_hqu.pdf) in last week's reco meeting, DeepAK8 is only somewhat slower than DeepDoubleB (which we currently don't include). It's actually faster than DeepFlavour, because there are more AK4 jets than AK8 jets.
* We're currently not running DeepAK8 or DeepFlavour for the Zinv clean jets, so look forward to an even slower TreeMaker once somebody decides they really want to use these variables for an analysis. (Maybe we can run just the decorrelated DeepAK8 for Zinv, since stealth stop doesn't use Zinv variables.)

### DIFF
--- a/Production/test/condorSub/.prodconfig
+++ b/Production/test/condorSub/.prodconfig
@@ -4,5 +4,4 @@ dir = .
 input = 2017B,2017C,2017D,2017E,2017F
 [caches]
 $CMSSW_BASE/src/NNKit/misc = 1
-$CMSSW_BASE/src/NNKit/data = 1
 $CMSSW_BASE/test = 1
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ The following installation instructions assume the user wants to process 2016 or
 wget https://raw.githubusercontent.com/TreeMaker/TreeMaker/Run2_2017/setup.sh
 chmod +x setup.sh
 ./setup.sh
-cd CMSSW_9_4_8/src/
+cd CMSSW_9_4_9/src/
 cmsenv
 cd TreeMaker/Production/test
 ```
--- a/Reflex/src/classes.h
+++ b/Reflex/src/classes.h
@@ -15,7 +15,6 @@ namespace {
     edm::Wrapper<std::vector<TLorentzVector> > wvlv;
     edm::Wrapper<std::vector<std::vector<TLorentzVector> > > wvvlv;
     edm::Wrapper<std::vector<std::vector<double> > > wvvd;
-    edm::Wrapper<std::vector<std::vector<int> > > wvvi;
     edm::Wrapper<std::vector<std::vector<bool> > > wvvb;
 	edm::Wrapper<edm::PtrVector<pat::PackedCandidate> > wrv2pp;
   };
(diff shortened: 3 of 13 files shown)


kpedro88:
I updated our fork of DeepAK8/NNKit with this commit (a few minor technical improvements):
https://gitlab.cern.ch/DeepAK8/NNKit/commit/8c636cc565fe569045e78cad986ed494f8e688d9

aperloff:
I thought without DeepAK8 we were up to 15 events/sec. Am I just woefully out of date? Is there some loss in speed due to the MET fix?

kpedro88:
There is a small loss due to the MET fix, but in general the 15 evt/sec number comes from ttbar relval without PU. Now I'm using the central GJet 2017 sample, which is more realistic.

Also, what appeared to be a minor improvement in DeepAK8 actually seems to have cut the time down to O(fastjet) instead of 16 times slower. I guess there are a lot of worthless jets in the latest miniAOD. So this puts us at 10 events/sec even while running DeepAK8 twice (full and decorrelated). I'm going to keep checking this...

kpedro88:
See https://github.com/cms-sw/cmssw/pull/20638 for info about low-pT AK8 jets in the latest miniAOD.

I am going to merge this PR, then make another one reorganizing all producers acting on AK8 jets to come after the GoodJetsProducer is run (which drops all of the useless jets).

PR #340, opened 2018-08-23
HEM 15/16 High Stats Samples

aperloff:
Add the files necessary to ntuplize the high stats HEM samples.


kpedro88:
also update readme w/ new scenario

aperloff:
Darn it! I always forget that one. Done now.

PR #342, opened 2018-08-27
2017 Data Lumi Calc

aperloff:
This PR is concerned with calculating the luminosity processed in making the 2017 data ntuples. It adds a dictionary of 2017 data samples, renames a 2016 dictionary, and allows for non-local versions of brilcalc to be used.


kpedro88:
the 2016 data is not prompt

aperloff:
Fixed

PR #347, opened 2018-09-07
Modifications to files needed during 2016 data job submission

aperloff:
I found at least one set of typos during job submission. We should probably leave this PR open until all or most of the jobs have finished running.


kpedro88:
@aperloff is this finished now?

aperloff:
Yes, this is finished.

PR #349, opened 2018-09-10
[Run2_2017] Pileup dataset checking for get_mcm.py

aperloff:
Update get_mcm.py to find and print the validity of the pileup dataset. If the pu dataset is incorrect the dataset is added to the unfinished cartegory and a notation is made. This commit also adds some additional samples to the default get_mcm dictionary.


kpedro88:
Let's add a third output file to store the samples that are finished, but have the wrong PU.

aperloff:
Okay, the invalid datasets now go to a third file. The unfinished file takes precedence, so unfinished datasets with invalid pu datasets go into the unfinished file.

PR #355, opened 2018-09-24
check for wrong cert at beginning of job

kpedro88:


aperloff:
Looks good to me. Hopefully all of those types of jobs have an identity like `*cmsgli*`.

PR #369, opened 2018-11-12
Updates to process old 2016 files

kpedro88:
Since the 2016 miniAODv3 campaign in 94X is not converging quickly, we need to be able to process the old 80X MC. A few physics objects need to be rerun to make the outputs correspond to what we have in 94X:
* DeepCSV tagger for AK4 jets (don't bother with DeepFlavour for now)
* Puppi + AK8 clustering + substructure for AK8 jets (80X uses PFCHS)
* BadChargedCandidate, BadPFMuon MET filters (were already rerun in our config)

There are a few other differences to handle as well:
* different era for electron post-reco tool
* `selectedPatTrigger` rather than `slimmedPatTrigger`

A new modifier `TM80X` is introduced to handle these changes that are specific to the 80X MC. The python file lists, scenarios, and unit tests for the old MC are restored. The miniAODv3 GT will be used.

Rerunning the objects listed above reduces the event rate to 7 ev/s. The second-largest contribution (after jet clustering) is from the calculation of ECFs. These could be temporarily disabled (to be considered in a subsequent PR).


kpedro88:
Validated by running on the exact same file in an old version of TreeMaker (Run2 branch, 8_0_30) and comparing quantities that aren't expected to change (e.g. NMuons, photon pT, gen variables). Jets and MET quantities change slightly due to updated JECs, etc.

PR #372, opened 2018-11-13
[Run2_2017] Add new features to the setup script.

aperloff:
Changes to the Run2_2017 setup script to allow for the user to specify a name and location for the CMSSW directory. The usual defaults are kept. This also allows the script to determine the appropriate slc version for the users computer system. These new features should be ported to the Run2_2018_prompt branch, eventually.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,26 +1,31 @@
 #!/bin/bash -e
 
+CMSSWVER=CMSSW_9_4_10
+FORK=TreeMaker
+BRANCH=Run2_2017
+ACCESS=ssh
+CORES=8
+NAME=""
+DIR="${PWD}"
+
 usage(){
 	EXIT=$1
 
 	echo "setup.sh [options]"
 	echo ""
-	echo "-f [fork]           clone from specified fork (default = TreeMaker)"
-	echo "-b [branch]         clone specified branch (default = Run2_2017)"
-	echo "-a [protocol]       use protocol to clone (default = ssh, alternative = https)"
-	echo "-j [cores]          run CMSSW compilation on # cores (default = 8)"
+	echo "-f [fork]           clone from specified fork (default = ${FORK})"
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -9,6 +9,8 @@ usage(){
 	echo "-b [branch]         clone specified branch (default = Run2_2017)"
 	echo "-a [protocol]       use protocol to clone (default = ssh, alternative = https)"
 	echo "-j [cores]          run CMSSW compilation on # cores (default = 8)"
+    echo "-n [name]           name of the CMSSW directory (default = CMSSW_9_4_10)"
```
kpedro88: this file uses tabs for indentation rather than 4 spaces, please make it consistent

setup.sh
```diff
@@ -47,11 +55,31 @@ else
 fi
 
 # get CMSSW release
-export SCRAM_ARCH=slc6_amd64_gcc630
+if [[ `uname -r` == *"el6"* ]]; then
+    SLC_VERSION="slc6"
+elif [[ `uname -r` == *"el7"* ]]; then
+    SLC_VERSION="slc7"
+else
+    echo "WARNING::Unknown SLC version. Defaulting to SLC7."
```
kpedro88: default should be sl6 as long as that is the CMS default

setup.sh
kpedro88: also, define the variable CMSSWVER before this function, and use it to print the default (instead of duplicating CMSSW_9_4_10)

setup.sh
aperloff: Fixed

setup.sh
aperloff: Fixed

kpedro88: by "make it consistent" I meant use tabs everywhere, not spaces

aperloff: Whoops. My bad. Fixed.

PR #381, opened 2018-11-26
Pdf updates

kpedro88:
* add simple analyzer `PdfFinder` to dump LHE headers
* add parameters for `PDFWeightProducer`
* store PS weights if available
* recalculate PDF weights if not present (currently only for SVJ signal)

Some details may need to change in the future, see: https://hypernews.cern.ch/HyperNews/CMS/get/generators/4156.html

### DIFF
--- a/TreeMaker/python/makeTreeFromMiniAOD_cff.py
+++ b/TreeMaker/python/makeTreeFromMiniAOD_cff.py
@@ -98,7 +98,10 @@ def makeTreeFromMiniAOD(self,process):
     ## ----------------------------------------------------------------------------------------------
     if self.geninfo and self.doPDFs:
         from TreeMaker.Utils.pdfweightproducer_cfi import PDFWeightProducer
-        process.PDFWeights = PDFWeightProducer.clone()
+        process.PDFWeights = PDFWeightProducer.clone(
+            recalculatePDFs = cms.bool(self.signal),
+            pdfSetName = cms.string("NNPDF31_lo_as_0130"),
+        )
         self.VectorDouble.extend(['PDFWeights:PDFweights','PDFWeights:ScaleWeights','PDFWeights:PSweights'])
 
     ## ----------------------------------------------------------------------------------------------
--- a/TreeMaker/python/makeTreeFromMiniAOD_cff.py
+++ b/TreeMaker/python/makeTreeFromMiniAOD_cff.py
@@ -99,7 +99,7 @@ def makeTreeFromMiniAOD(self,process):
     if self.geninfo and self.doPDFs:
         from TreeMaker.Utils.pdfweightproducer_cfi import PDFWeightProducer
         process.PDFWeights = PDFWeightProducer.clone()
-        self.VectorDouble.extend(['PDFWeights:PDFweights','PDFWeights:ScaleWeights'])
+        self.VectorDouble.extend(['PDFWeights:PDFweights','PDFWeights:ScaleWeights','PDFWeights:PSweights'])
 
     ## ----------------------------------------------------------------------------------------------
     ## GenHT for stitching together MC samples
--- a/TreeMaker/python/makeTreeFromMiniAOD_cff.py
+++ b/TreeMaker/python/makeTreeFromMiniAOD_cff.py
@@ -97,9 +97,9 @@ def makeTreeFromMiniAOD(self,process):
     ## PDF weights for PDF systematics
     ## ----------------------------------------------------------------------------------------------
     if self.geninfo and self.doPDFs:
-        process.PDFWeights = cms.EDProducer('PDFWeightProducer')
+        from TreeMaker.Utils.pdfweightproducer_cfi import PDFWeightProducer
+        process.PDFWeights = PDFWeightProducer.clone()
         self.VectorDouble.extend(['PDFWeights:PDFweights','PDFWeights:ScaleWeights'])
-        self.VectorInt.extend(['PDFWeights:PDFids'])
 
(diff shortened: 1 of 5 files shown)


TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -97,9 +97,12 @@ def makeTreeFromMiniAOD(self,process):
     ## PDF weights for PDF systematics
     ## ----------------------------------------------------------------------------------------------
     if self.geninfo and self.doPDFs:
-        process.PDFWeights = cms.EDProducer('PDFWeightProducer')
-        self.VectorDouble.extend(['PDFWeights:PDFweights','PDFWeights:ScaleWeights'])
-        self.VectorInt.extend(['PDFWeights:PDFids'])
+        from TreeMaker.Utils.pdfweightproducer_cfi import PDFWeightProducer
+        process.PDFWeights = PDFWeightProducer.clone(
+            recalculatePDFs = cms.bool(self.signal),
```
aperloff, line 102: Is this what specifies that this should just be done for the SVJ signals? It seems like it will include all signals.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -97,9 +97,12 @@ def makeTreeFromMiniAOD(self,process):
...
+            pdfSetName = cms.string("NNPDF31_lo_as_0130"),
```
aperloff, line 103: Do you know how often this name will change? Is there any way to get it from the tune name or the sample automatically?

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -97,9 +97,12 @@ def makeTreeFromMiniAOD(self,process):
     ## PDF weights for PDF systematics
     ## ----------------------------------------------------------------------------------------------
     if self.geninfo and self.doPDFs:
-        process.PDFWeights = cms.EDProducer('PDFWeightProducer')
-        self.VectorDouble.extend(['PDFWeights:PDFweights','PDFWeights:ScaleWeights'])
-        self.VectorInt.extend(['PDFWeights:PDFids'])
```
aperloff, line 102: What happened to this line? Is it not needed anymore?

Utils/src/PDFWeightProducer.cc
```diff
@@ -63,57 +123,84 @@ void PDFWeightProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::Ev
...
+
+      int id1 = genHandle->pdf()->id.first;
+      double x1 = genHandle->pdf()->x.first;
+      double pdf1 = genHandle->pdf()->xPDF.first;
+
+      int id2 = genHandle->pdf()->id.second;
+      double x2 = genHandle->pdf()->x.second;
+      double pdf2 = genHandle->pdf()->xPDF.second;
+      if(pdf1 <= 0 && pdf2 <= 0) {
+         std::lock_guard<std::mutex> lock(mutex_);
```
aperloff, line 175: What is the mutex_ doing? I understand that you are probably making the code thread safe, but I see a lock and no corresponding unlock. FYI, I have never use a mutex and have no idea what I'm talking about. Just trying to understand the code a little better.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
kpedro88: The other signals should have the pdf weights already stored. It only tries to recalculate them if they aren't stored *and* if this flag is true. We can double-check this later.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
kpedro88: Unfortunately not, as far as I can tell. See my long hypernews post for more details. The only way to get this info is parsing the massive LHE header block, and that seems not entirely reliable to me (in addition to being overly complicated).

kpedro88: This is the one used in TuneCP2.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
kpedro88: Look below, we're now importing it from a cfi file (to keep track of the parameters).

Utils/src/PDFWeightProducer.cc
kpedro88: `mutex_` is a static member variable of the class. If it is locked in one thread, other threads will wait to execute this part of the code until it is unlocked. The lock is created within the scope of the if statement, so once it goes out of scope and destructs, the mutex is unlocked.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
aperloff: I meant the saving of the PDFids. Is saving that not important anymore? Is that the same as the pdfSetName? Is see nothing in the cfi that says "ID" to me.

Utils/src/PDFWeightProducer.cc
aperloff: Got it. Thanks for the explanation.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
kpedro88: Ah, no, I don't think it matters. The ID was basically just the index of the weight in the vector, so we don't really need to keep that info separately.

PR #382, opened 2018-11-28
Updates for 2018 data in CMSSW_10_2_X

kpedro88:
The relevant updates from the branch `Run2_2018_prompt` are propagated back to `Run2_2017`.

The setup script is updated to support *both* CMSSW_9_4_X and CMSSW_10_2_X in the same branch. If this starts to cause problems, we will make a separate branch. So far, it seems to work.

The higher-stats 2018B HEM samples (JetHT and EGamma) are added for processing. The ntuples will go under Run2ProductionV16a.

Also a few minor tweaks to scripts.


aperloff:
Looks good to me.

PR #390, opened 2018-12-18
[Run2_2017] Getting ready to run on the 2018 data

aperloff:
This PR updates the Run2_2017 branch for processing the 2018 data. It includes the necessary dictionaries, file lists, scenarios, unit tests, and readme. There is also a small update to get_mcm.py.


kpedro88:
Is all of the 2018 re-reco done now? I though some samples were still processing.

aperloff:
They paused the processing and told us to run on what was there. The time to completion is at least several months.

kpedro88:
Okay... this is going to be a huge pain in the ass for file lists, because when we regenerate them, the order will change.

aperloff:
Nothing I can do about that. We could institute some sorting.

kpedro88:
The file lists are sorted, but the missing files could appear anywhere. I guess we'll burn that bridge after we jump off it.

PR #404, opened 2019-01-23
[Run2_2017] Prepare for V16 FastSim Ntuple Production

aperloff:
This PR updates the scenarios, file lists, and dictionaries needed to produce the Summer16MiniAODv3 and Fall17MiniAODv2 fastsim ntuples. The PR is currently missing an update to the JEC file, which probably needs to be updated. Please carefully check the new scenarios.


aperloff:
It appears @kpedro88 already did this, so this PR was unnecessary.

PR #406, opened 2019-01-31
[Run2_2017] Update GJets_DR-0p4_HT Samples

aperloff:
This PR updates the RunIIFall17MiniAODv2 dictionaries to include new GJets_DR-0p4_HT samples and remove the old, invalid samples. The updates include:
1) File lists
2) condorSub dictionaries
3) getWeightProducer lines


kpedro88:
I don't think we need to call them v2, just use the usual name.

aperloff:
It wasn't my choice. The name of the sample was changed in McM, so this is what get_py.py decided to name the files. I guess I can override it if you'd like. What say you?

kpedro88:
Ah, it's directly in the first part of the dataset name, not from the second part. I guess we should leave it as is.

PR #414, opened 2019-02-11
[Run2_2017] Update QCD_HT File Lists For New Premixed Samples

aperloff:
Add the file lists for the two VALID QCD_HT samples with the new PU premixing.


aperloff:
I will update the weight producer lines in a second PR.

PR #417, opened 2019-02-22
[Run2_2017] RunIIAutumn18MiniAOD MC Reload

aperloff:
This PR reloads the dictionaries and settings needed to run on the 2018 MC samples. It includes the following changes:

1) The 2018 MC scenarios
2) An update of the 2018 MC JEC
3) Update RunIIAutumn18MiniAOD file lists, including one RunIIAutumn18MiniAOD-HEM file list
4) Update PFJetID selection to be more flexible. Also update the 2018 PFJetID definition
5) Update the WeightProducer to include the gen event info weight in the event weight when running on a flat sample
6) The QCD flat sample cross section
7) The condorSub dictionaries for the 2018 MC samples, including the neff dictionary

Note: This is probably not the last reload of the 2018 MC.
Note: This does not include the fix to the PFJetID which would allow for backing out the JER scale factor. That should be put into TreeMaker before the V17 production.
Note: The JEC will be updated again shortly when the JERC group releases the L2L3Residual corrections (for data).


Production/test/condorSub/dict_RunIIAutumn18MiniAOD_dyjets.py
```diff
@@ -3,5 +3,11 @@
     "samples": [
         ['RunIIAutumn18MiniAOD.DYJetsToLL_M-50_HT-100to200_TuneCP5_PSweights_13TeV-madgraphMLM-pythia8'],
         ['RunIIAutumn18MiniAOD.DYJetsToLL_M-50_TuneCP5_13TeV-madgraphMLM-pythia8_ext1'],
+        ['RunIIAutumn18MiniAOD.DYJetsToLL_M-50_HT-200to400_TuneCP5_PSweights_13TeV-madgraphMLM-pythia8'],
+        ['RunIIAutumn18MiniAOD.DYJetsToLL_M-50_HT-400to600_TuneCP5_PSweights_13TeV-madgraphMLM-pythia8'],
+        ['RunIIAutumn18MiniAOD.DYJetsToLL_M-50_HT-600to800_TuneCP5_PSweights_13TeV-madgraphMLM-pythia8'],
+        ['RunIIAutumn18MiniAOD.DYJetsToLL_M-50_HT-800to1200_TuneCP5_PSweights_13TeV-madgraphMLM-pythia8'],
+        ['RunIIAutumn18MiniAOD.DYJetsToLL_M-50_TuneCP1_13TeV-madgraphMLM-pythia8'],
```
kpedro88: do we really want TuneCP1?

TreeMaker/python/makeJetVars.py
```diff
@@ -130,9 +135,10 @@ def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInput
     BTagsDeepCSV = btagint.clone(
         JetTag       = HTJetsTag,
         BTagInputTag = cms.string('pfDeepCSVDiscriminatorsJetTags:BvsAll'),
-        BTagCutValue = cms.double(0.6324)
+        BTagCutValue = cms.double(0.6321)
```
kpedro88: I think we should not make changes to the 2016 settings until we prepare for the V17 ntuples.

Utils/src/GoodJetsProducer.cc
```diff
@@ -45,85 +46,87 @@
...
+private:
+	bool filter(edm::StreamID, edm::Event&, const edm::EventSetup&) const override;
+
+	// ----------member data ---------------------------
+	edm::InputTag JetTag_;
+	std::vector<edm::InputTag> SkipTag_;
+	edm::EDGetTokenT<edm::View<pat::Jet>> JetTok_;
+	std::vector<edm::EDGetTokenT<edm::View<reco::Candidate>>> SkipTok_;
+	double maxEta_;
+	//double maxNeutralFraction_, maxNeutralFractionHE_, minNeutralFractionHF_, maxPhotonFraction_, minPhotonFractionHE_, maxPhotonFractionHE_, maxPhotonFractionHF_, minChargedFraction_, maxChargedEMFraction_;
```
kpedro88: remove commented-out code (everywhere in this file)

Utils/src/GoodJetsProducer.cc
```diff
@@ -45,85 +46,87 @@
...
+	JetTok_ = consumes<edm::View<pat::Jet>>(JetTag_);
+	for(auto& tag: SkipTag_) SkipTok_.push_back(consumes<edm::View<reco::Candidate>>(tag));
+
+	// check the contents of the varnames_ vector to make sure it contains only allowed values
+	vector<string> varoptions = {"neutralEmEnergyFraction","neutralHadronEnergyFraction","neutralMultiplicity",
+								 "chargedEmEnergyFraction","chargedHadronEnergyFraction","chargedMultiplicity",
+								 "nconstituents","muonEnergyFraction","nef","nhf","nm","cef","chf","cm","nc","mf"};
+	for(auto v : varnames_) {
+		if (std::find(varoptions.begin(), varoptions.end(), v) == varoptions.end()) {
+			std::cout << "Input error: variable name " << v << " unknown" << endl;
```
kpedro88: cout is not allowed, please use MessageLogger

Utils/src/GoodJetsProducer.cc
```diff
@@ -45,85 +46,87 @@
...
+
+	JetTok_ = consumes<edm::View<pat::Jet>>(JetTag_);
+	for(auto& tag: SkipTag_) SkipTok_.push_back(consumes<edm::View<reco::Candidate>>(tag));
+
+	// check the contents of the varnames_ vector to make sure it contains only allowed values
+	vector<string> varoptions = {"neutralEmEnergyFraction","neutralHadronEnergyFraction","neutralMultiplicity",
+								 "chargedEmEnergyFraction","chargedHadronEnergyFraction","chargedMultiplicity",
+								 "nconstituents","muonEnergyFraction","nef","nhf","nm","cef","chf","cm","nc","mf"};
+	for(auto v : varnames_) {
+		if (std::find(varoptions.begin(), varoptions.end(), v) == varoptions.end()) {
```
kpedro88: make `varoptions` a `set` (or `unordered_set`) to avoid N linear searches

Utils/src/GoodJetsProducer.cc
```diff
@@ -133,122 +136,113 @@ GoodJetsProducer::~GoodJetsProducer()
...
+				}
+				else leptonMask->push_back(false);
+			}
+			else leptonMask->push_back(false);
+
+			//calculate the PFJetID decision
+			bool good = true;
+			double varval = 0.0;
+			for(unsigned v=0; v<varnames_.size(); ++v) {
+				if(varnames_[v]=="neutralHadronEnergyFraction" || varnames_[v]=="nhf")      varval = iJet.neutralHadronEnergyFraction();//gives raw energy in the denominator
```
kpedro88: make the available var names into enums to avoid N*M string comparisons
(or go for the full factory approach of JetProperties, might be overkill)

kpedro88:
8. Update the DeepCSV working point for 2018

Production/test/condorSub/dict_RunIIAutumn18MiniAOD_dyjets.py
aperloff: We certainly do not. This will be fixed in my next commit.

TreeMaker/python/makeJetVars.py
aperloff: I will revert this back to 0.6324, but I've made a note about this at https://twiki.cern.ch/twiki/bin/view/CMS/RA2b13TeVProduction#Low_priority.

Utils/src/GoodJetsProducer.cc
aperloff: Whoops. I meant to do this before. Fixed now.

Utils/src/GoodJetsProducer.cc
aperloff: Fixed

Utils/src/GoodJetsProducer.cc
aperloff: Good idea. Done.

Utils/src/GoodJetsProducer.cc
aperloff: Still working on this one.

aperloff: Take a look at this. I switched to using enums, though I'm not sure if this is exactly what you were thinking. I didn't go the full factory approach because It seemed like overkill to build a class for each type of energy fraction/multiplicity. It sounds interesting, but I would need help to implement it.

TreeMaker/python/makeJetVars.py
kpedro88: I had already made a note "Update DeepCSV WP for 2016"

Utils/src/GoodJetsProducer.cc
kpedro88: Yes, that is what I was thinking. Fine for now

Utils/src/GoodJetsProducer.cc
```diff
@@ -20,6 +20,8 @@
 
 // system include files
 #include <memory>
+#include <algorithm>
```
kpedro88: these includes are no longer needed

kpedro88:
Have you tested the GoodJetProducer updates to make sure it gives identical results to the old version?

aperloff:
**TLDR:**
I have tested the GoodJetProducer. It doesn't produce identical results, which is good because the old version had a bug which is fixed in the new version.

**The long version:**
In the old method, the cuts were hard coded (i.e. value < maxValueHE). Therefore if we want to turn off a cut we would set maxValueHE to 1.0. A safer value would have been 9999, but 1.0 was what was in the configuration file. Because of the problem with not backing out the JER, the value can be greater than 1.0. Therefore, that jet would fail the PFJetID even though it wasn't supposed to fail on that specific cut. About 5-8% of events (mostly low pT jets) were affected by the problem of not properly turning off the PFJetID cuts.

In the new method, the jet is only checked against a cut if there is either a lower or upper bound. Cuts which are supposed to be turned off either don't get put into the vector, or can be turned off by setting the boundary value equal to -1.0. Therefore, this problem of false positives is removed.

PR #424, opened 2019-03-03
Remove JER smearing from energy fractions and MHT fix

kpedro88:


aperloff:
Looks good to me. Did you check that if you draw, for example, the chargeHadronEnergyFraction that none of the values go beyond 1? This was happening more for low pt jets with some regularity, but you would need to do it for more than 100 events, so I would plan on testing it with 1000 or 10000 events.

kpedro88:
I had only tested with 100 events. I will run a larger test.

kpedro88:
Confirmed with 10K events that all fractions are < 1.

aperloff:
Okay I approve. Or in the parlance of cms-sw/cmssw...+1

PR #425, opened 2019-03-04
Jet ID changes for PUPPI

kpedro88:
Also, added rho variable for PU studies.

NB: 2018 PUPPI-specific Jet ID settings are not available yet.


aperloff:
Looks good to me.

PR #432, opened 2019-03-11
improve NeffFinder

kpedro88:
Store the `NeffFinder` results in a histogram (easier to track than log files).

The output ROOT file from `NeffFinder` is now called `NeffInfo_...`. The histogram with neff info is also called `NeffInfo_...`. The `TrueNumInteractions_...` histogram is also present in that file.

The four bins in the `NeffInfo_...` histogram correspond to neff (pos-neg), pos, neg, and tot (pos+neg).

I confirmed that the content in the histogram is the same as what `getResults.py` would return after aggregating the log files.


aperloff:
1) Can the bins be labeled? It would make it easier if we don't have to go back to the code to find the bin-to-count mapping.
2) Do you have an example file I can look at?

kpedro88:
ROOT has done backward-incompatible things with labeled bins before, so I would prefer to avoid that.

Example files in /store/user/lpcsusyhad/SusyRA2Analysis2015/Autumn18NeffTest/

aperloff:
You should put the bin mapping in the README then.

PR #437, opened 2019-03-25
[Run2_2017_V17] Save minimal set of gen particles

aperloff:
This PR adds a new way to store a collection of gen particles. Previously we were storing a very large set of particles. Now we will only save a small subset of important decays (i.e. tops, W, Z, H, etc.) We will also save SUSY particles and particles important for semi-visible jets. A flag to switch between the two methods has been added to makeTreeFromMiniAOD.

Currently this has been tested using semi-leptonic TTJets decays. The command to do this was:
`cmsRun runMakeTreeFromMiniAOD_cfg.py scenario=Fall17 inputFilesConfig=Fall17.TTJets_SingleLeptFromT_TuneCP5_13TeV-madgraphMLM-pythia8 nstart=0 nfiles=1 outfile=Fall17.TTJets_SingleLeptFromT_TuneCP5_13TeV-madgraphMLM-pythia8 numevents=100 saveMinimalGenParticles=True`

### DIFF
--- a/TreeMaker/python/makeTreeFromMiniAOD_cff.py
+++ b/TreeMaker/python/makeTreeFromMiniAOD_cff.py
@@ -137,23 +137,45 @@ def makeTreeFromMiniAOD(self,process):
     ## GenParticles
     ## ----------------------------------------------------------------------------------------------
     if self.geninfo:
-        process.genParticles = cms.EDProducer("GenParticlesProducer",
-            genCollection = cms.InputTag("prunedGenParticles"),
-            debug = cms.bool(False),
-            childIds = cms.vint32(1,2,3,4,5,11,12,13,14,15,16,22),
-            parentIds = cms.vint32(
-                1,2,6,23,24,25,
-                1000021,1000022,1000023,1000024,1000025,1000035,1000037,1000039,
-                1000001,1000002,1000003,1000004,1000005,1000006,
-                2000001,2000002,2000003,2000004,2000005,2000006,
-                4900021,4900023,4900101,4900102,4900111,4900113,4900211,4900213,51,52,53,
-                5000001,5000002,
-            ),
-            keepIds = cms.vint32(6,23,24,25),
-            keepFirst = cms.bool(True),
-        )
-        # store gluons for signals with Higgs
-        if "T5qqqqZH" in process.source.fileNames[0]: process.genParticles.childIds.append(21)
+        if self.saveMinimalGenParticles:
+            process.genParticles = cms.EDProducer("GenParticlesProducer",
+                genCollection = cms.InputTag("prunedGenParticles"),
+                debug = cms.bool(False),
+                # Particles we want to save from the decay chain of the tops
+                childIds = cms.vint32(1,2,3,4,5,11,12,13,14,15,16,24),
+                # Particles we want to save the last copy from the hard scatter
+                parentIds = cms.vint32(
+                    6,22,23,24,25,
+                    1000021,1000022,1000023,1000024,1000025,1000035,1000037,1000039,
+                    1000001,1000002,1000003,1000004,1000005,1000006,
+                    2000001,2000002,2000003,2000004,2000005,2000006,
+                    4900021,4900023,4900101,4900102,4900111,4900113,4900211,4900213,51,52,53,
+                    5000001,5000002,
+                ),
+                # Other settings
+                keepIds = cms.vint32(),
+                keepFirst = cms.bool(False),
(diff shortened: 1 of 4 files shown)


kpedro88:
Testing:

Using the sample `Fall17.TTJets_SingleLeptFromT_TuneCP5_13TeV-madgraphMLM-pythia8`

A handful of events produce warnings like this:
```
%MSG
%MSG-i TreeMaker:  GenParticlesProducer:genParticles 25-Mar-2019 13:41:26 CDT  Run: 1 Event: 49842344
WARNING The last particle in the chain (1) does not have a status that corresponds to its type (status=23)
%MSG
%MSG-i TreeMaker:  GenParticlesProducer:genParticles 25-Mar-2019 13:41:48 CDT  Run: 1 Event: 49842754
WARNING The last particle in the chain (5) does not have a status that corresponds to its type (status=73)
%MSG
```

Without this option, `GenParticles` is the largest single branch and the second-largest family of branches (1.58 and 1.74 kb/evt, resp.). The total event size is 8.39 kb/evt.

With this option, `GenParticles` goes down to the third-largest branch and fourth-largest family of branches (0.33 and 0.35 kb/evt, resp.). The total event size is 7.03 kb/evt.

This represents a reduction in size of 16.2%. The corresponding savings in the V16 MC ntuples would be roughly 8TB (assuming the same reduction for all samples).

Comments:
* enable the option by default
* add the option to the list in README.md

aperloff:
The option has bee added to the README and the default is now set to True.

aperloff:
I have added status code 23 and 73 to the list of possible final state quark status codes.

kpedro88:
For `Fall17.GJets_HT-400To600_TuneCP5_13TeV-madgraphMLM-pythia8`, the reduction is 10% (7.18 kb/evt -> 6.43 kb/evt).

Utils/src/GenParticlesProducer.cc
```diff
@@ -99,7 +113,164 @@ void GenParticlesProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+    ==========================================
+    From a semi-leptonic TTJets sample:
+    -----------------------------------
+    1. tops start with status 22 and end with 62
+        any time there is one daughter it is simply radiating. when there are two daughters that should be the final top
+    2. b start with status 23 and end with 71/72 (ignore any gluon radiation)
+    3. W start with status 22, can immediately decay or go to status 51/52
+        immediately decays to status 23 which then decay to status 1/2 (leptonic) or 71 (hadronic)
+    */
+    if(((last->pdgId() >= down && last->pdgId() <= bottom) && (last->status()!=23 && last->status()!=51 && last->status()!=52 && last->status()!=71 && last->status()!=72 && last->status()!=73)) ||
```
kpedro88: 1. shouldn't all these be `abs(pdgId)`?
2. probably easier to put `pdgId` and `status` into temporary vars
3. for the list of quark status codes, make it a sorted vector and use `std::binary_search()`
4. are these status codes checked anywhere else, or is it just for this warning?

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -137,23 +137,45 @@ def makeTreeFromMiniAOD(self,process):
...
+                    4900021,4900023,4900101,4900102,4900111,4900113,4900211,4900213,51,52,53,
+                    5000001,5000002,
+                ),
+                keepIds = cms.vint32(6,23,24,25),
+                keepFirst = cms.bool(True),
+                keepMinimal = cms.bool(False),
+            )
+            # store gluons for signals with Higgs
+            if "T5qqqqZH" in process.source.fileNames[0]: process.genParticles.childIds.append(21)
         self.VectorTLorentzVector.append("genParticles(GenParticles)")
```
kpedro88, line 179: The branch `TTFlag` can be removed in the minimal case

Utils/src/GenParticlesProducer.cc
aperloff: 1. Done
2. Done
3. Done
4. The status codes are only checked for that warning message. I thought about making it not store that depth if the codes weren't right, but since I'm still empirically finding new, acceptable code, that seemed prone to error.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
aperloff: Done

aperloff:
Retested the code to see that it is removing the TTFlag branch and the updated status code warning is working. Everything looks good from my side.

PR #443, opened 2019-03-29
[Run2_2017] Update XS and k-Factors

aperloff:
- Update the year-by-year XS and k-factor values. Values taken from XSDB, https://twiki.cern.ch/twiki/bin/viewauth/CMS/SummaryTable1G25ns, and GenXSecAnalyzer.

- Make sure the k-factors are propagated to the weightProducer XS value.

PR #444, opened 2019-03-29
use 2016 xsecs for all 13 TeV (for now)

kpedro88:
Since #443 is deferred for later, make sure every sample gets assigned a cross section.


kpedro88:
NB, typo subsequently fixed in https://github.com/TreeMaker/TreeMaker/commit/e84fbcfeffac463b906a83d9e80fdf2f98f121db

PR #446, opened 2019-04-16
[Run2_2017] Add the Summer16v3/Fall17 T1ttbb FastSim signal samples.

aperloff:
Add the file list and condorSub entry for the Fall17Fast T1ttbb signal sample.


kpedro88:
there should also be a Summer16v3 fastsim sample for this process

aperloff:
I've added in the Summer16v3 sample and I'm going to change the name of the PR to be more inclusive.

PR #447, opened 2019-04-16
[Run2_2017] Add to V17 FastSim samples and cleanup various issues

aperloff:
Some tops decay to a b and W without and intermediate step. This means that the only top(s) in the event will have status 22 and not status 62. I've added status 22 as an option for the final event status.


aperloff:
I didn't initially mean to combine two commits, but I don't see a problem with making all of these changes here. I'm just going to change the title of the PR.

Production/test/dataSamples_Run2018_17Sep2018.py
```diff
@@ -16,7 +16,8 @@
     ["SingleMuon_2018C", "Run2018C-17Sep2018-v1.SingleMuon_*"],
     # 2018D datasets: PromptReco
     ["EGamma_2018D", "Run2018D-PromptReco-v2.EGamma_*"],
```
kpedro88: this line should also be removed?

PR #448, opened 2019-04-30
Updates for signal samples (primarily SVJ)

kpedro88:
The SVJ-oriented changes are:
* allow disabling PDF weight normalization
* store SVJ `GenParticle` chains (with a hack to keep first copies of HV quarks)
* prepare for upcoming SVJ signal scans

Some of these changes are not quite backward compatible for V17:
* new `SignalParameters` branch generalized to handle any type of signal scan (`SusyMotherMass` and `SusyLSPMass` branches kept for compatibility, plan to remove them after V17 finishes)
* `GenParticles_ParentIdx` branch filled, when possible, for "individual" particles kept in the minimal scheme
* `GenParticle` chain added to minimal scheme for SUSY signals

but the impact is minor, so we will go ahead.

attn @aperloff for `GenParticleProducer` changes


aperloff:
Why did you add the parent_list to Parent_vec part. Storing the (last copy) parent positions was already being taken care of?

kpedro88:
Parent indices were only being stored for particles in specific chains. Individual particles never had their parent indices computed and stored. In principle, an individual particle's parent could have been stored by a chain, so we need to know every possible parent position in order to do the complete check.

aperloff:
okay

PR #463, opened 2019-09-25
[Run2_2017] Update The DeepDoubleB Tagger

aperloff:
Add in updated versions of the DeepDoubleB jet tagger. This is the officially recommended version, as opposed to the slightly older pre-release version which was being used before. This PR maintains both versions of the tagger for comparison and study.


kpedro88:
for production i would prefer to get rid of the old one if it is no longer needed/useful

aperloff:
The old discriminators have been removed

PR #470, opened 2020-01-07
[Run2_2017] Update the JECs used by the AK8 subjets and softdrop mass

aperloff:
Currently the subjets stored in the ntuples and used for the softdrop mass do not use any JECs. This PR will store the subjets with JECs and will use those to compute the softdrop mass. These same subjets will also be propagated to the other properties which make use of subjets.

Along the way, this PR grew to add the subjet JEC to the tree and to add a debugging option to store the subjet response.

The code has been checked and the JECs are updated when using the new code (if it's commented out the original subjets from the MiniAOD will be used). The attached plot shows what happens to the response with and without the new code.

[SubjetJECUpdate.pdf](https://github.com/TreeMaker/TreeMaker/files/4032913/SubjetJECUpdate.pdf)


aperloff:
One question, do we need to resort the subjets in case the updated JEC cause them to no longer be in descending order of pt?

kpedro88:
updateJetCollection should re-sort them automatically

PR #472, opened 2020-02-14
[Run2_2017] Changes to how the trigger names are stored

aperloff:
Add the ability to set a docstring/title for each branch. This mapping between InputTags and titles will be used to store the trigger names as a comma separated list in the branch title, rather than a vector of strings for each event. We end up storing the list three times, once per trigger branch, but that is still far less than the once per event we were doing before.

These modifications have been successfully tested using unit test 1 and 100 events. The log file size went from 63K to 75K, which is due to printing the InputTag to title association. This is a big string in the case of the trigger names. The number of branches went from 424 to 423, as expected.

The ROOT file size went from 879K to 754K, but this doesn't tell the entire story as the header size should stay roughly the same, ignoring the title strings, but the per event size will go down significantly. For the original method, the sum of the sizes of the compressed (uncompressed) branches* is 805.34 KB (2067.67 KB). For the new method this goes down to 680.64 KB (1645.88 KB). The average per event size of the tree* goes from 8.05 KB (20.68 KB) to 6.81 KB (16.46 KB), or a ~15.4% reduction in event size.

The TriggerNames branch took 113.46 KB (421.80 KB), or 1.13 KB (4.22 KB) per event. The fact that the compressed values don't make up the difference between the old and new methods can be explained by different compression ratios in the branches and other, more in-depth ROOT optimizations. 

* Data and offsets, not including any key headers.

### DIFF
--- a/TreeMaker/interface/TreeMaker.h
+++ b/TreeMaker/interface/TreeMaker.h
@@ -18,6 +18,7 @@
 
 //ROOT headers
 #include "TString.h"
+#include "TBranch.h"
 #include "TTree.h"
 #include <TFile.h>
 #include "TLorentzVector.h"
@@ -71,6 +72,7 @@ class TreeMaker : public edm::one::EDAnalyzer<edm::one::SharedResources> {
 		UInt_t lumiBlockNum{};
 		ULong64_t evtNum{};
 		vector<TreeObjectBase*> variables;
+		map<string,string> titleMap;
 };
 
 //base class for tree objects
@@ -78,7 +80,7 @@ class TreeObjectBase {
 	public:
 		//constructor
 		TreeObjectBase() : tempFull(""), branchType("") {}
-		TreeObjectBase(string tempFull_) : tempFull(tempFull_), nameInTree(tempFull_), tagName(tempFull_), tree(nullptr) {}
+		TreeObjectBase(string tempFull_, string title_ = "") : tempFull(tempFull_), nameInTree(tempFull_), tagName(tempFull_), title(title_), tree(nullptr) {}
 		//destructor
 		virtual ~TreeObjectBase() {}
 		//functions
@@ -86,6 +88,7 @@ class TreeObjectBase {
 		virtual void Initialize(map<string,unsigned>& nameCache, edm::ConsumesCollector && iC, stringstream& message) {}
 		virtual void SetTree(TTree* tree_) { tree = tree_; }
 		virtual void AddBranch() {}
+		virtual void AddTitle() { if(branch) branch->SetTitle(title.c_str()); }
 		virtual void SetDefault() {}
 		virtual void FillTree(const edm::Event& iEvent) {}
 		
@@ -108,9 +111,10 @@ class TreeObjectBase {
 		
 	protected:
 		//member variables
-		string tempFull, nameInTree, tagName, branchType;
+		string tempFull, nameInTree, tagName, branchType, title;
 		TTree* tree{};
 		edm::InputTag tag;
+		TBranch* branch{};
 };
 
 //comparator (case-insensitive sort)
@@ -132,7 +136,7 @@ class TreeObject : public TreeObjectBase {
 	public:
 		//constructor
 		TreeObject() : TreeObjectBase() {}
-		TreeObject(string tempFull_) : TreeObjectBase(tempFull_) {}
+		TreeObject(string tempFull_, string title_) : TreeObjectBase(tempFull_,title_) {}
 		//destructor
 		~TreeObject() override {}
 		//functions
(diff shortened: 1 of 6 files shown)


TreeMaker/interface/TreeMaker.h
```diff
@@ -71,21 +72,23 @@ class TreeMaker : public edm::one::EDAnalyzer<edm::one::SharedResources> {
...
-		TreeObjectBase(string tempFull_) : tempFull(tempFull_), nameInTree(tempFull_), tagName(tempFull_), tree(nullptr) {}
+		TreeObjectBase(string tempFull_, string title_ = "") : tempFull(tempFull_), nameInTree(tempFull_), tagName(tempFull_), title(title_), tree(nullptr) {}
 		//destructor
 		virtual ~TreeObjectBase() {}
 		//functions
 		virtual string GetNameInTree() const { return nameInTree; }
 		virtual void Initialize(map<string,unsigned>& nameCache, edm::ConsumesCollector && iC, stringstream& message) {}
 		virtual void SetTree(TTree* tree_) { tree = tree_; }
 		virtual void AddBranch() {}
+		virtual void AddTitle() { if(branch) branch->SetTitle(title.c_str()); }
```
kpedro88: I think this should be `if(branch and !title.empty())`

TreeMaker/src/TreeMaker.cc
```diff
@@ -48,10 +48,32 @@ TreeMaker::TreeMaker(const edm::ParameterSet& iConfig) :
...
+	// parse the TitleMap
 	stringstream skipMessage;
 	stringstream message;
+	set<string> titleSet;
+	vector<string> titleVector = iConfig.getParameter< vector<string> >("TitleMap");
+	message << "TitleMap:\n";
+	for(unsigned i = 0; i < titleVector.size(); i+=2){
+		string variable = titleVector.at(i);
+		string title = titleVector.at(i+1);
+		message << "full name:" << variable << " -> title: " << title << "\n";
```
kpedro88: since we already print the trigger list once in the logs, we probably don't need to print it 3 more times... maybe we can add another TreeMaker option like debugTitles and only print this if it is enabled?

TreeMaker/src/TreeMaker.cc
```diff
@@ -48,10 +48,32 @@ TreeMaker::TreeMaker(const edm::ParameterSet& iConfig) :
...
+		//check for an exact repeat of an existing name
+		//otherwise add an entry to the titleMap
+		if (titleSet.find(variable)!=titleSet.end()){
+			if(skipMessage.str().empty()) skipMessage << "Skipping repeated (branch,title) pairs:\n";
```
kpedro88: technically this doesn't check for a repeated pair, but just a repeated key (branch name)

TreeMaker/src/TreeMaker.cc
```diff
@@ -48,10 +48,32 @@ TreeMaker::TreeMaker(const edm::ParameterSet& iConfig) :
...
-	//loop over all var type names to initialize TreeObjects
-	set<string> nameSet;
+
+	// parse the TitleMap
 	stringstream skipMessage;
 	stringstream message;
+	set<string> titleSet;
+	vector<string> titleVector = iConfig.getParameter< vector<string> >("TitleMap");
+	message << "TitleMap:\n";
+	for(unsigned i = 0; i < titleVector.size(); i+=2){
```
kpedro88, line 65: I would check for an odd-size vector before this, and probably just throw an exception in that case
(it would be a bit clearer then just getting an out of range exception from `vector.at()`)

TreeMaker/src/TreeMaker.cc
kpedro88: also, prepending `"Skipping repeated branches:\n"` to `skipMessage` on L119 no longer makes sense, should probably take a similar approach down below

TreeMaker/interface/TreeMaker.h
aperloff: Someone has been writing too much python, but yes, you are right.

kpedro88: that's perfectly valid c++ :p

TreeMaker/src/TreeMaker.cc
aperloff: Done

TreeMaker/src/TreeMaker.cc
aperloff: I tried to fix both the branch/title message and the printing of the original skipMessage. Let me know what you think.

TreeMaker/src/TreeMaker.cc
aperloff: Done

TreeMaker/src/TreeMaker.cc
```diff
@@ -48,39 +48,77 @@ TreeMaker::TreeMaker(const edm::ParameterSet& iConfig) :
...
-	set<string> nameSet;
+	debugTitles = iConfig.getParameter<bool>("debugTitles");
+
+	// parse the TitleMap
 	stringstream skipMessage;
 	stringstream message;
+	set<string> titleSet;
+	vector<string> titleVector = iConfig.getParameter< vector<string> >("TitleMap");
+	if(titleVector.size()%2 != 0) {
+		cms::Exception ex("LogicError", "The vector containing the InputTags and titles is odd, indicating that there is not a 1-to-1 pairing");
```
kpedro88: "is odd" -> "has an odd number of entries"

TreeMaker/src/TreeMaker.cc
kpedro88: Looks fine to me now, thanks

TreeMaker/src/TreeMaker.cc
aperloff: Fixed

PR #473, opened 2020-02-19
[Run2_2017] Save track information

aperloff:
Changes to allow for the saving of track information which is stored in the MiniAOD samples. This will add a lot of information to the ntuples, so it is turned off by default. This was tested using the unit test command:

```bash
python unitTest.py test=1 run=true command="emerging=True"
```


Reflex/src/classes_def.xml
```diff
@@ -1,11 +1,18 @@
...
+  <class name="std::vector<std::vector<math::XYZVector> >"/>
+  <class name="std::vector<math::XYZPoint>"/>
+  <class name="std::vector<std::vector<math::XYZPoint> >"/>
   <class name="std::vector<pat::Jet>"/>
   <class name="std::vector<std::vector<pat::Jet> >"/>
+  <class name="std::vector<std::vector<bool> >"/>
   <class name="edm::PtrVector<pat::PackedCandidate>"/>
   <class name="edm::Wrapper<std::vector<TLorentzVector> >"/>
   <class name="edm::Wrapper<std::vector<std::vector<TLorentzVector> > >"/>
+  <class name="edm::Wrapper<std::vector<std::vector<math::XYZVector> > >"/>
```
kpedro88, line 14: were you going to remove these as redundant?

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -1115,6 +1117,52 @@ def makeTreeFromMiniAOD(self,process):
...
+            'trackFilter:trksnormalizedchi2(Tracks_normalizedChi2)',
+            'trackFilter:trkspterror(Tracks_ptError)',
+            'trackFilter:trksetaerror(Tracks_etaError)',
+            'trackFilter:trksphierror(Tracks_phiError)',
+            'trackFilter:trksqoverperror(Tracks_qoverpError)',
+            'trackFilter:trksip2d(Tracks_IP2DPV0)',
+            'trackFilter:trksip2dsig(Tracks_IP2dSigPV0)',
+            'trackFilter:trksip3d(Tracks_IP3DPV0)',
+            'trackFilter:trksip3dsig(Tracks_IP3DSigPV0)',
+            'trackFilter:pfcandsdzassociatedpv(PFCandidates_dzAssociatedPV)',
```
kpedro88: This saves some quantities associated with PFCandidates (distinct from tracks), but doesn't save the PFCandidate 4-vectors. Would those also be needed? (Leading to another large space increase...)

TreeMaker/python/maker.py
```diff
@@ -101,6 +102,8 @@ def __init__(self,parameters):
         self.VarsInt                    = cms.vstring()
         self.VarsBool                   = cms.vstring()
         self.VectorTLorentzVector       = cms.vstring()
+        self.VectorXYZVector            = cms.vstring()
+        self.VectorXYZPoint             = cms.vstring()
```
kpedro88, line 108: you should add the other new branch types here also

TreeMaker/python/maker.py
```diff
@@ -124,6 +127,7 @@ def printSetup(self):
         print " storing hadtau variables: "+str(self.hadtau)+" w/ reclustering "+str(self.hadtaurecluster)
         print " storing Zinv variables: "+str(self.doZinv)
         print " storing semi-visible jet variables: "+str(self.semivisible)
+        print " storing emerging jet variables: "+str(self.semivisible)
```
kpedro88: wrong var in str()

Utils/src/CandidateTrackFilter.cc
```diff
@@ -0,0 +1,380 @@
...
+		trks_dxypv->push_back(track.dxy());
+		trks_dxyerrorpv->push_back(track.dxyError());
+		trks_normalizedchi2->push_back(track.normalizedChi2());
+		trks_pterror->push_back(track.ptError());
+		trks_etaerror->push_back(track.etaError());
+		trks_phierror->push_back(track.phiError());
+		trks_qoverperror->push_back(track.qoverpError());
+		trks_found->push_back(track.found());
+		trks_lost->push_back(track.lost());
+		// track quality enum options: http://cmsdoxygen.web.cern.ch/cmsdoxygen/CMSSW_10_2_11_patch1/doc/html/d8/df2/classreco_1_1TrackBase.html#aeee12ec6a3ea0d65caa2695c84ab25d5
```
kpedro88: would prefer to replace doxygen links w/ github links

Utils/src/CandidateTrackFilter.cc
```diff
@@ -0,0 +1,380 @@
...
+	infos.put(iEvent);
+	return result;
+}
+
+// ------------ other methods  ------------
+bool CandidateTrackFilter::filterOnTrack(const pat::PackedCandidate & pfCand, const reco::Track & track) const {
+	//-------------------------------------------------------------------------------------
+	// only store PFCandidate values if pt >= minPt and |eta| <= maxEta_
+	//-------------------------------------------------------------------------------------
+	if ( (track.pt() < minPt_) || (fabs(track.eta()) > maxEta_) ) return false;
```
kpedro88: use `std::abs()` rather than `fabs()`

Utils/src/CandidateTrackFilter.cc
```diff
@@ -0,0 +1,380 @@
...
+		// fill the track values and PFCandidate values we'd like to save
+		//-------------------------------------------------------------------------------------
+		auto transientTrack = ttBuilder->build(track);
+		infos.fill(primaryVertex,lostTrack,track,transientTrack,false);
+	}
+
+	//-------------------------------------------------------------------------------------
+	// put track/PFCandidate values back into event
+	//-------------------------------------------------------------------------------------
+	bool result = infos.trks->empty();
```
kpedro88: I am concerned about letting this "filter" ever actually reject events. result should be configured as true by default, with a separate option to enable event filtering if it is desired for some reason.

Utils/src/CandidateTrackFilter.cc
```diff
@@ -0,0 +1,380 @@
...
+bool CandidateTrackFilter::filterOnTrack(const pat::PackedCandidate & pfCand, const reco::Track & track) const {
+	//-------------------------------------------------------------------------------------
+	// only store PFCandidate values if pt >= minPt and |eta| <= maxEta_
+	//-------------------------------------------------------------------------------------
+	if ( (track.pt() < minPt_) || (fabs(track.eta()) > maxEta_) ) return false;
+
+	//-------------------------------------------------------------------------------------
+	// cut on impact parameter quantities
+	//-------------------------------------------------------------------------------------
+	if ( (std::abs(track.dz()) > maxdz_) || (std::abs(track.dxy()) > maxdxy_) ) return false;
```
kpedro88: this whole function could just be:
```C++
return track.pt() >= minPt_ and std::abs(track.eta()) <= maxEta_ and ...
```

Reflex/src/classes_def.xml
aperloff: The vector<vector<object>> weren't redundant, only the vector<object> ones.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
aperloff: Technically the pT of the PFCandidate doesn't need to match that of the track, but according to the MiniAOD twiki that is rare. Most often the information is the same, so the only thing you would be adding is an energy measurement. Do you think this is needed?

TreeMaker/python/maker.py
aperloff: What other ones? vector<vector<object>>?

TreeMaker/python/maker.py
aperloff: Whoops! Now fixed.

Utils/src/CandidateTrackFilter.cc
aperloff: Yes, but I wanted to leave room to add other cuts/selections is need be. Putting it all into one line is not by preferred approach. Are you insisting on that format?

Utils/src/CandidateTrackFilter.cc
aperloff: Done

Utils/src/CandidateTrackFilter.cc
aperloff: Done

Utils/src/CandidateTrackFilter.cc
aperloff: Implemented

TreeMaker/python/makeTreeFromMiniAOD_cff.py
kpedro88: So the vectors starting with "Track_" and the vectors starting with "PFCandidate_" are supposed to be the same size? This breaks our naming convention.

TreeMaker/python/maker.py
kpedro88: `VarsXYZVector`, etc. Every new line in `treeMaker.py`

Utils/src/CandidateTrackFilter.cc
kpedro88: `if bool return bool` is a redundant pattern. You can write a single return statement on multiple lines if desired.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
aperloff: They should be the same size by construction. How does this break the naming convention?

kpedro88: Branches with different names are not expected to have the same size.

Utils/src/CandidateTrackFilter.cc
aperloff: Okay, this has been changed.

TreeMaker/python/maker.py
aperloff: Fixed

TreeMaker/python/makeTreeFromMiniAOD_cff.py
aperloff: Fixed

aperloff:
All of the comments from the review should be addressed now.

README.md
```diff
@@ -225,6 +225,7 @@ Brief explanation of the options in [makeTree.py](./TreeMaker/python/makeTree.py
 * `doZinv`: switch to enable the Z->invisible background estimation processes (default=True)
 * `systematics`: switch to enable JEC- and JER-related systematics (default=True)
 * `semivisible`: switch to enable variables for semi-visible jets (default=True)
+* `emerging`: switch to enable variables for emerging jets (default=True)
```
kpedro88: default should be false?

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -38,6 +38,8 @@ def makeTreeFromMiniAOD(self,process):
         VarsInt                    = self.VarsInt,
         VarsBool                   = self.VarsBool,
         VectorTLorentzVector       = self.VectorTLorentzVector,
+        VectorXYZVector            = self.VectorXYZVector,
+        VectorXYZPoint             = self.VectorXYZPoint,
```
kpedro88, line 44: for completeness the new branch lists should also be added here

Utils/src/CandidateTrackFilter.cc
```diff
@@ -0,0 +1,366 @@
+//-----------------------------------------------------------------------------------------
```
kpedro88: this large totally irrelevant block comment can be removed

Utils/src/CandidateTrackFilter.cc
```diff
@@ -0,0 +1,366 @@
...
+#include "DataFormats/ParticleFlowCandidate/interface/PFCandidateFwd.h"
+#include "DataFormats/PatCandidates/interface/PackedCandidate.h"
+#include "DataFormats/VertexReco/interface/VertexFwd.h"
+
+#include "TrackingTools/IPTools/interface/IPTools.h"
+#include "TrackingTools/TransientTrack/interface/TransientTrackBuilder.h"
+#include "TrackingTools/Records/interface/TransientTrackRecord.h"
+#include "MagneticField/Engine/interface/MagneticField.h"
+
+#include "CommonTools/UtilAlgos/interface/TFileService.h"
```
kpedro88: unneeded include can be removed

Utils/src/CandidateTrackFilter.cc
```diff
@@ -0,0 +1,366 @@
...
+
+#include "Math/VectorUtil.h"
+#include "TMath.h"
+#include "TLorentzVector.h"
+
+//using namespace reco;
```
kpedro88: delete commented-out code

README.md
aperloff: Done

TreeMaker/python/makeTreeFromMiniAOD_cff.py
aperloff: Done

Utils/src/CandidateTrackFilter.cc
aperloff: Done, if all you meant was the TFileService.h.

Utils/src/CandidateTrackFilter.cc
aperloff: Done

Utils/src/CandidateTrackFilter.cc
aperloff: Done

PR #475, opened 2020-02-24
[Run2_2017] Save information for primary vertices

aperloff:
Changes the PrimaryVerticesInt producer to save the good vertex and all primary vertices ints at the same time. Also allows use to save the primary vertex information for all vertices, marking off if it was included in the collection of good vertices. Without saving any emerging jets information, the per event size is 6.81 KiB. After saving just the track information this goes to 24.02 KiB. After the addition of the vertices, the size again grows to 30.42 KiB.

### DIFF
--- a/TreeMaker/python/makeTreeFromMiniAOD_cff.py
+++ b/TreeMaker/python/makeTreeFromMiniAOD_cff.py
@@ -1122,7 +1122,22 @@ def makeTreeFromMiniAOD(self,process):
     ## Emerging jets
     ## ----------------------------------------------------------------------------------------------
     if self.emerging:
-        self.VarsInt.extend(['primaryVertices:_______(Vertices_)'])
+        self.VectorXYZPoint.extend(['primaryVertices:vtxposition(PrimaryVertices_position)'])
+        self.VectorDouble.extend([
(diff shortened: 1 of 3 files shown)


TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -126,15 +126,12 @@ def makeTreeFromMiniAOD(self,process):
         filter = cms.bool(False)
     )
     from TreeMaker.Utils.primaryvertices_cfi import primaryvertices
-    process.NVtx = primaryvertices.clone(
-        VertexCollection  = cms.InputTag('goodVertices'),
+    process.primaryVertices = primaryvertices.clone(
+        VertexCollection     = cms.InputTag('offlineSlimmedPrimaryVertices'),
+        GoodVertexCollection = cms.InputTag('goodVertices'),
+        saveVertices         = cms.bool(True if self.emerging else False),
```
kpedro88: `cms.bool(self.emerging)`

aperloff: Fixed

PR #479, opened 2020-03-03
[Run2_2017] Modify the way PFJetID is saved

aperloff:
Instead of saving the PFJetID as true if a jet is matched to a lepton, store the real PFJetID value and rely on checking the LeptonMask. The previous method has its historical roots in the way the LostLepton background estimation was done. Post-RA2/b, it makes more sense to save the true PFJetID for all jets and rely more heavily on checking the LeptonMask if need be.

This PR shouldn't have any effect on the speed of production or size of the ntuples. The effect was checked on one event which was known to have a lepton jet. The event was (run,lumi,event)=(1,206,160306) in the Summer16v3.SingleLeptFromT_TuneCUETP8M1v1 dataset. The only jet which was matched to a lepton had it's PFJetID switch from "true" to "false," which was the behavior we were after. It should fail the PFJetID because it was in the barrel region, but only had one constituent, when the requirement is to have more than one.


aperloff:
Feel free to comment, but please wait to merge this until the CU/UCSB synchronization exercise is complete.

PR #488, opened 2020-03-31
[Run2_2017] Additional information needed for the HH+MET Synchronization Exercise

aperloff:
This PR adds some additional information needed for the HH+MET synchronization exercise. In particular it adds the following variables to the PFCandidates:
1. The PF Relative Isolation (DR=0.3) for only the charged component
2. The PF Relative Isolation (DR=0.3) for all components
3. The dxy impact parameter
These branches are only added if the 'debugtracks' option is set to True.

Adding these branches necessitated adding a new isolation calculator to the SUSYIsolation class. There is one change (L89 in Utils/interface/get_isolation_activity.h) that needs some special review. This as added based upon https://github.com/cms-sw/cmssw/blob/56d3b7f00bac17989db6efb3efaf8e1ba31f2dfe/PhysicsTools/PatAlgos/plugins/PATIsolatedTrackProducer.cc#L554, which is the producer that makes the 'isolatedTrack' collection for MiniAOD. I think the change is well motivated.

This PR has been unit tested and compared to 37 events (50 PF candidates) from NanoAOD. The comparisons show that the new isolation values match up to 4 decimal places.


Utils/interface/get_isolation_activity.h
```diff
@@ -126,12 +126,41 @@ class SUSYIsolation {
...
+                      double r_pfiso=0.3, double ptThresh=0.0) const
+  {
+    pf_iso_all = 0.0;
+    pf_iso_chg = 0.0;
+    const auto& ptcl = pfcands->at(pfcand_idx);
+    if (ptcl.p4().Pt()<5.) {
+    pf_iso_all = 99999.;
+    pf_iso_chg = 99999.;
+    return;
+  }
```
kpedro88: fix indentation of this block

Utils/interface/get_isolation_activity.h
```diff
@@ -86,7 +86,7 @@ class SUSYIsolation {
         }
       }
       //////////////////  CHARGED from PV  /////////////////////////
-    } else if (pfc.fromPV()>1){
+    } else if (pfc.fromPV()>1 || abs(pfc.dz()) < 0.1){
```
kpedro88: `std::abs()`

aperloff: Done

Utils/interface/get_isolation_activity.h
aperloff: Done

PR #489, opened 2020-04-02
[Run2_2017] Add missing EMJ collections

aperloff:
Add some track collections which were missing for the EMJ analysis. To do this required adding some ultra-legacy (UL) eras as one of the collections is only available in UL.

Also update the part of the unit test code which handles arbitrary options/commands passed to TreeMaker. Before, it was only allowing one (option=value) pair. Now it can handle many space separated pairs. Outside of that pattern, it returns to the previous way of passing these options.


TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -1094,7 +1094,10 @@ def makeTreeFromMiniAOD(self,process):
                 vertexInputTag    = cms.InputTag("goodVertices"),
                 pfCandidatesTag   = cms.InputTag("packedPFCandidates"),
                 lostTracksTag     = cms.InputTag("lostTracks"),
+                lostEleTracksTag  = cms.InputTag("lostTracks","eleTracks"),
         )
+        if TMeras.TMUL2016.isChosen() or TMeras.TMUL2017.isChosen() or TMeras.TMUL2018.isChosen():
```
kpedro88: better way:
```python
(TMeras.TMUL2016 | TMeras.TMUL2017 | TMeras.TMUL2018).toModify(process.trackFilter,
    displacedStandAloneMuonsTag = cms.InputTag("displacedStandAloneMuons")
)
```

TreeMaker/python/TMEras.py
```diff
@@ -10,6 +10,9 @@ def __init__(self):
             'TM2016_80X',
             'TM2017',
             'TM2018',
+            'TMUL2016',
```
kpedro88, line 13: these won't actually work until we upgrade TreeMaker to be compatible with 10_6_X, right?

aperloff: Well, there's no scenario that references them right now, so it's not something I've actually tested. You're right that UL is 10_6_X and that TreeMaker probably needs to be updated to work with that. Still, there's no harm in having these eras available so that we can slowly add these sorts of features.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
aperloff: adopted

PR #490, opened 2020-04-02
[Run2_2017] Add a new Docker image for ServiceX

aperloff:
Add the automatic creation of a new Docker image for ServiceX. This creation will take place after the base image for TreeMaker and simultaneous to the integration test. The reason this image is needed is to add some customization to the python packages and environment variables.

In doing this work, several updated were made to the GitHub actions workflow.
1. Publishing is turned off for all users except 'TreeMaker'. While this isn't totally secure, it prevents innocent users from accidentally overwriting the images if they happen to have the DockerHub secrets.
2. The integration-test and service-x workflows won't run if the user isn't 'TreeMaker'. As the image won't be updated unless the user is merging in the Run2_2017 branch after a PR and the users account can't publish the resulting images, it doesn't make sense to go through the effort for them to run a second , but unpublished, integration test.

Because of the changes mentioned above, I was unable to test to ServiceX workflow in it's entirety. I was able to test that if one uses the treemaker/treemaker:Runw_2017-latest image, that the .github/ServiceX/scripts/setup.sh works. The only part that is untested is the .github/ServiceX/scripts/setup-root.sh part.

### DIFF
--- a/.github/ServiceX/python/transformer.py
+++ b/.github/ServiceX/python/transformer.py
@@ -64,7 +64,7 @@ def callback(channel, method, properties, body):
 
     servicex.post_status_update(file_id=_file_id,
                                 status_code="start",
-                                info="xAOD Transformer")
+                                info="MiniAOD Transformer")
 
     tick = time.time()
     file_done = False
@@ -73,7 +73,7 @@ def callback(channel, method, properties, body):
         try:
             # Do the transform
             root_file = _file_path.replace('/', ':')
-            output_path = '/home/atlas/' + root_file
+            output_path = '/home/cmsuser/' + root_file
             transform_single_file(_file_path, output_path, servicex)
 
             tock = time.time()
@@ -123,8 +123,6 @@ def callback(channel, method, properties, body):
 
 def transform_single_file(file_path, output_path, servicex=None):
     print("Transforming a single path: " + str(file_path) + " into " + output_path)
-    # os.system("voms-proxy-info --all")
-    #r = os.system('bash /generated/runner.sh -r -d ' + file_path + ' -o ' + output_path +  '| tee log.txt')
     r = os.system('python ' + os.environ['CMSSW_BASE'] + '/TreeMaker/Production/test/unitTest.py test=0 scenario=Summer16v3 dataset=file:' + str(file_path) + ' name=' + str(output_path) + ' run=True fork=False log=True')
     reason_bad = None
     if r != 0:
@@ -166,7 +164,7 @@ def compile_code():
 
 
 if __name__ == "__main__":
-    parser = TransformerArgumentParser(description="xAOD CPP Transformer")
+    parser = TransformerArgumentParser(description="MiniAOD CPP Transformer")
     args = parser.parse_args()
 
     kafka_brokers = TransformerArgumentParser.extract_kafka_brokers(args.brokerlist)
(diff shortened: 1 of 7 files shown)


.github/ServiceX/python/transformer.py
```diff
@@ -0,0 +1,187 @@
...
+                                info="xAOD Transformer")
+
+    tick = time.time()
+    file_done = False
+    file_retries = 0
+    while not file_done:
+        try:
+            # Do the transform
+            root_file = _file_path.replace('/', ':')
+            output_path = '/home/atlas/' + root_file
```
kpedro88: do we really have to use `/home/atlas/`?

.github/ServiceX/python/transformer.py
```diff
@@ -0,0 +1,187 @@
...
+                                            status_code="retry",
+                                            info="Try: " + str(file_retries) +
+                                                 " error: " + str(error)[0:1024])
+
+    channel.basic_ack(delivery_tag=method.delivery_tag)
+
+
+def transform_single_file(file_path, output_path, servicex=None):
+    print("Transforming a single path: " + str(file_path) + " into " + output_path)
+    # os.system("voms-proxy-info --all")
```
kpedro88: delete commented-out code

.github/ServiceX/python/transformer.py
```diff
@@ -0,0 +1,187 @@
...
+    # this file.
+    r = os.system('bash /generated/runner.sh -c | tee log.txt')
+    if r != 0:
+        with open('log.txt', 'r') as f:
+            errors = f.read()
+            raise RuntimeError("Unable to compile the code - error return: " + str(r)+ 'errors: \n' + errors)
+
+
+if __name__ == "__main__":
+    parser = TransformerArgumentParser(description="xAOD CPP Transformer")
```
kpedro88: xAOD -> miniAOD

.github/ServiceX/scripts/proxy-exporter.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+#sudo mkdir -p /etc/grid-security
```
kpedro88, line 2: delete commented-out code

.github/ServiceX/scripts/proxy-exporter.sh
```diff
@@ -0,0 +1,13 @@
...
+#sudo chown cmsuser /etc/grid-security
+
+
+while true; do
+    #sudo cp /etc/grid-security-ro/x509up /etc/grid-security
```
kpedro88, line 7: delete commented-out code

.github/ServiceX/python/transformer.py
aperloff: Will change this to `/home/cmsuser/`

.github/ServiceX/python/transformer.py
aperloff: Done

.github/ServiceX/python/transformer.py
aperloff: Done

.github/ServiceX/scripts/proxy-exporter.sh
aperloff: That is there on purpose. It will be needed eventually, but for this first test @mweinberg2718 and I are sidestepping anything having to do with `sudo` or grid proxies.

.github/ServiceX/scripts/proxy-exporter.sh
aperloff: Same as the comment above.

PR #492, opened 2020-04-03
[Run2_2017] Transform nested vector branches to flattend+offset vector branches

aperloff:
Allow the transformation of vector<vector<T>> branches to be saved as flattened vector<T> branches with an associated vector<int> branch of offsets.

Two unit tests were performed, one keeping the nested vector format and the other flattening the vectors. Flattening the vectors has the effect of decreasing the ntuple size by a little under 1% (0.88% in our 100 event test).

I have not tested the speed change.

### DIFF
--- a/TreeMaker/interface/TreeMaker.h
+++ b/TreeMaker/interface/TreeMaker.h
@@ -359,20 +359,19 @@ class TreeRecoCand : public TreeObject<vector<TLorentzVector> > {
 		vector<double> pt, eta, phi, energy;
 };
 
-//derived version of vector<vector<T>>
-//with switch for vector<T> values and vector<int> offsets instead
-template <template<typename...> class R=std::vector, typename Top = R<R<double>>, typename Sub = typename Top::value_type>
+// Derived version of vector<vector<T>> with switch for vector<T> values and vector<int> offsets instead
+template <typename Base = double, typename Sub = std::vector<Base>, typename Top = std::vector<Sub>> 
 class TreeNestedVector : public TreeObject<Top> {
 	public:
-		//constructor
+		// Constructor
 		TreeNestedVector() : TreeObject<Top>() {}
-		TreeNestedVector(string type_, string tempFull_, string title_="", bool nestedVectors_=true) : TreeObject<Top>(tempFull_,title_), type(type_), nestedVectors(nestedVectors_) {}
-		//destructor
+		TreeNestedVector(string tempFull_, string title_="", bool nestedVectors_=true) : TreeObject<Top>(tempFull_,title_), nestedVectors(nestedVectors_) {}
+		// Destructor
 		~TreeNestedVector() override {}
 		
-		//functions
-		//From: https://stackoverflow.com/questions/17294629/merging-flattening-sub-vectors-into-a-single-vector-c-converting-2d-to-1d
-		void flatten(Top const& all, R<typename Sub::value_type> &accum, R<int> &offsets) {
+		// Functions
+		// From: https://stackoverflow.com/questions/17294629/merging-flattening-sub-vectors-into-a-single-vector-c-converting-2d-to-1d
+		void flatten(Top const& all, Sub &accum, vector<int> &offsets) {
 			// Don't store any offsets if there are no sub-vectors
 			if (all.size() > 0)	offsets.insert(std::end(offsets),0);
 			else 				return;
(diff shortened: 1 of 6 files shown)


TreeMaker/interface/TreeMaker.h
```diff
@@ -356,3 +358,83 @@ class TreeRecoCand : public TreeObject<vector<TLorentzVector> > {
 		bool doLorentz{};
 		vector<double> pt, eta, phi, energy;
 };
+
+//derived version of vector<vector<T>>
+//with switch for vector<T> values and vector<int> offsets instead
+template <template<typename...> class R=std::vector, typename Top = R<R<double>>, typename Sub = typename Top::value_type>
```
kpedro88: in what case would R not be std::vector?

kpedro88: wouldn't we prefer to just make the inner type a template, and then build up Sub and Top from that?

TreeMaker/interface/TreeMaker.h
```diff
@@ -356,3 +358,83 @@ class TreeRecoCand : public TreeObject<vector<TLorentzVector> > {
...
+			else{
+				values.clear();
+				offsets.clear();
+			}
+		}
+
+	protected:
+		//member variables
+		edm::EDGetTokenT<Top> tok;
+		string type{};
```
kpedro88: default initialization with `{}` is not necessary for strings (string class is automatically initialized to an empty string)

TreeMaker/python/maker.py
```diff
@@ -146,6 +147,7 @@ def printSetup(self):
         print " Storing a minimal set of GenParticles: "+str(self.saveMinimalGenParticles)
         print " Storing the GenTops: "+str(self.saveGenTops)
         print " Saving the MT2 variable: "+str(self.doMT2)
+        print " Saving nested vectors as vector<vector<T>>: "+str(self.nestedVectors)
```
kpedro88: i would make this message conditional:
```python
if self.nestedVectors: print " Saving nested vectors as vector<vector<T>>"
else: print " Saving nested vectors as vector<T> + vector<int>"
```

TreeMaker/interface/TreeMaker.h
aperloff: 1) I guess in our case we should restrict R to be std::vector. I took this code from somewhere else and if you look at the flatten function it technically doesn't require the outer object to be a std::vector. Do you want me to work on removing R?
2) Doing (1) would make this possible, but is it really necessary? Seems to be roughly equivalent code.

TreeMaker/interface/TreeMaker.h
aperloff: Changed

TreeMaker/interface/TreeMaker.h
kpedro88: I think it's preferable to make the code as simple as possible (or in other words, only as complex as necessary). It's simpler to specify one parameter, the inner type, and then just typedef Sub and Top as vectors of that inner type. The use of value_type is clever, but unnecessarily so.

TreeMaker/python/maker.py
aperloff: changed

TreeMaker/interface/TreeMaker.h
aperloff: Okay, the new commit should make things simpler.

TreeMaker/interface/TreeMaker.h
```diff
@@ -356,3 +358,110 @@ class TreeRecoCand : public TreeObject<vector<TLorentzVector> > {
 		bool doLorentz{};
 		vector<double> pt, eta, phi, energy;
 };
+
+// Derived version of vector<vector<T>> with switch for vector<T> values and vector<int> offsets instead
+template <typename Base = double, typename Sub = std::vector<Base>, typename Top = std::vector<Sub>>
```
kpedro88: in what case would we want Sub *not* to be `std::vector<Base>`?

I think Sub and Top should not be template parameters, but typedefs inside the class:
```c++
typedef Sub std::vector<Base>;
typedef Top std::vector<Sub>;
```

kpedro88: this means you'll have to specify ` : public TreeObject<std::vector<std::vector<Base>>>` in the class definition, but oh well

aperloff: Done

PR #498, opened 2020-04-09
[Run2_2017] Add option for a new output folder structure

aperloff:
This PR adds the option to store the output in a folder structure using the pattern \<output_dir>/\<era>/\<dataset>/\<filename>.root, rather than storing all of the files in the single <output_dir> directory. This change had been tested. This PR depends on https://github.com/kpedro88/CondorProduction/pull/7.


aperloff:
I have run two additional tests since moving the structuredOutput function to step2.sh, one with the `-f` flag and one without. Both tests behaved as expected.

PR #501, opened 2020-04-20
[Run2_2017] Additional tag-and-probe branches

aperloff:
Add dxy and pfRelIso03chg branches to the tag-and-proble collections. Also, reorder some of the python lines for adding collections.


aperloff:
Due to an oddity in the way ROOT compresses TFiles this change actually reduces the size of the event by 1%.

PR #502, opened 2020-04-21
[Run2_2017] Add a new workflow for ServiceX

aperloff:
Add new workflow for building images compatible with ServiceX. This workflow mirrors the repo to CERN GitLab and then triggers a CI job, which builds the images. The image built by this workflow distinguishes itself by having CMSSW installed inside the image, not needing CVMFS.

### DIFF
--- a/.github/ServiceX/Dockerfile
+++ b/.github/ServiceX/Dockerfile
@@ -1,0 +1,39 @@
+# Make the base image configurable:
+ARG BASEIMAGE=gitlab-registry.cern.ch/treemaker/cmssw-docker/cmssw:CMSSW_10_2_21-2020-04-07-fc019a29
+
+# Set up the CMSSW base:
+FROM ${BASEIMAGE}
+
+ARG BUILD_DATE
+ARG VCS_REF
+ARG VCS_URL
+ARG VERSION
+LABEL   org.label-schema.build-date=$BUILD_DATE \
+        org.label-schema.name="TreeMaker Docker image for ServiceX" \
+        org.label-schema.description="Provide completely offline-runnable CMSSW images with the TreeMaker and ServiceX dependencies pre-installed." \
+        org.label-schema.url="https://github.com/TreeMaker/TreeMaker" \
+        org.label-schema.vcs-ref=$VCS_REF \
+        org.label-schema.vcs-url=$VCS_URL \
+        org.label-schema.vendor="FNAL" \
+        org.label-schema.version=$VERSION \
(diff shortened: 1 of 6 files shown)


setup.sh
```diff
@@ -149,3 +171,5 @@ scram b -j ${CORES}
 # extra setup
 cd TreeMaker/Production/test/condorSub/
 python $CMSSW_BASE/src/Condor/Production/python/linkScripts.py
+
+set +x
```
kpedro88, line 175: should this be inside `if [[ "${DEBUG}" == "true" ]]` ?

.github/ServiceX/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+        org.label-schema.vcs-ref=$VCS_REF \
+        org.label-schema.vcs-url=$VCS_URL \
+        org.label-schema.vendor="FNAL" \
+        org.label-schema.version=$VERSION \
+        org.label-schema.schema-version="1.0"
+
+USER    cmsusr
+WORKDIR /home/cmsusr
+
+ARG CMSSW_VERSION=CMSSW_10_2_21
```
kpedro88, line 24: in the future, we'll have to remember to do a global find/replace whenever changing the CMSSW version...

setup.sh
```diff
@@ -22,13 +23,14 @@ usage(){
 	echo "-j [cores]          run CMSSW compilation on # cores (default = ${CORES})"
 	echo "-n [name]           name of the CMSSW directory (default = ${CMSSWVER})"
 	echo "-d [dir]            project installation area for the CMSSW directory (default = ${DIR})"
+	echo "-D                  print additional debug statements (default = ${DEBUG})"
```
kpedro88, line 26: add to readme?

setup.sh
aperloff: I can put it there if you'd like, but it doesn't really matter. If the default is `set +x` then doing it again won't make a difference.

.github/ServiceX/Dockerfile
aperloff: True. It's getting more complicated. The one you're pointing to is a build_arg, so it can be overridden by the .gitlab-ci.yml file, which is where the change ultimately needs to be. We also have a somewhat built-in safety mechanism here because we'd need to build a standalone CMSSW version for any new release.

setup.sh
aperloff: Done. I also added some other undocumented options.

PR #504, opened 2020-04-25
[Run2_2017] Update the lepton isolation branchs for consistency

aperloff:
The lepton isolation values saved to the tree were not consistent with the values used to determine the Lepton_passIso flags. While the passIso branches were produced in the LeptonProducer, the isolation values were created by the IsolationProducer. These two producers were accidentally using different settings for the lepton EA values. The IsolationProducer settings were out of date at this point.

Having these two producers was an unnecessary complication. Therefore, this PR removed the IsolationProducer and instead returns the isolation from the LeptonProducer.


aperloff:
Do not merge yet. We are conferring with the EGamma POG to see if there should be different EA values used for 2016 94X MC and 80X legacy MC. While I haven't found a recommendation on how to handle these on the EGamma twiki [1], we can look at what the NanoAOD is using [2]. The NanoAOD uses a different set of EA values for these MC eras.

[1]
https://twiki.cern.ch/twiki/bin/viewauth/CMS/CutBasedElectronIdentificationRun2
[2]
https://github.com/cms-sw/cmssw/blob/CMSSW_10_2_X/PhysicsTools/NanoAOD/python/electrons_cff.py#L113-L126

kpedro88:
ideally we're not relying on 80X samples any more for analysis, but if we can keep supporting it just in case, we might as well

Utils/src/TrackIsolationFilter.cc
```diff
@@ -319,6 +360,14 @@ void TrackIsolationFilter::GetTrkIso(edm::Handle<edm::View<pat::PackedCandidate>
 	activity = activity*invpt;
 }
 
+bool TrackIsolationFilter::hasLeptonMatch(const std::vector<reco::CandidatePtr>& leptonPfCands, const reco::CandidatePtr&  candPtr) const {
+	// Inspired by: https://github.com/cms-sw/cmssw/blob/424ad43856c2c739d702c240187401b7b08566ea/PhysicsTools/PatAlgos/plugins/IsolatedTrackCleaner.cc#L48-L49
+	if (std::binary_search(leptonPfCands.begin(), leptonPfCands.end(), candPtr)) {
```
kpedro88: this can just be:
```
return std::binary_search(leptonPfCands.begin(), leptonPfCands.end(), candPtr);
```

aperloff: Changed

PR #505, opened 2020-04-28
[Run2_2017] Update the cross sections and k-factors for V18

aperloff:
** DO NOT MERGE YET **

This PR updates the year-by-year XS and k-factor values. Values taken from XSDB, https://twiki.cern.ch/twiki/bin/viewauth/CMS/SummaryTable1G25ns, and GenXSecAnalyzer. 

One thing that still needs to be done is a second pass through to run the GenXSecAnalyzer on the 2018 samples that say "Same as 2017". This will be to confirm that those values really are the same.


aperloff:
The last of the 2018 XS have been updated. Please review.

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
-            "CrossSection" : XSValues(XS_13TeV=831.76),
+            "CrossSection" : XSValues(
+                XS_13TeV=831.76, XSSource_13TeV="XSDB (NNLO)",
+            ),
         },
-        # ttbar single lep & dilep xsecs scaled by PDG BR, assume t=tbar (hadronic: 377.96)
         "TTJets_SingleLeptFromT" : {
-            "CrossSection" : XSValues(XS_13TeV=182.72),
+            "CrossSection" : XSValues(
+                XS_13TeV=182.72, XSSource_13TeV="PDG xsec scales by PDG BR (t=tbar) - http://pdg.lbl.gov/2019/reviews/rpp2018-rev-top-quark.pdf",
```
kpedro88: scales by -> scaled by
(also in next few lines)

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+                XS_2017=0.7040, XSSource_2017="XSDB (LO)",
+                XS_2018=0.7028, XSSource_2018="GenXSecAnalyzer",
+            ),
         },
         "TTGamma_Dilept" : {
-            "CrossSection" : XSValues(XS_13TeV=0.5804),
+            "CrossSection" : XSValues(
+                XS_2016=0.6343, XSSource_2016="XSDB (NLO)",
+                XS_2017=0.5804, XSSource_2017="XSDB (LO)",
+                XS_2018=1.495,  XSSource_2018="GenXSecAnalyzer",
```
kpedro88: do we understand why this is x3 from 2017?

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+                kFactor_2017=1.162, kFactorSource_2017="XSDB NNLO/LO=61526.7/52940",
+                kFactor_2018=1.164, kFactorSource_2018="XSDB NNLO/LO=61526.7/52850",
+            ),
         },
         "WJetsToLNu_HT-2500ToInf" : {
-            "CrossSection" : XSValues(XS_13TeV=0.03891),
-            "kFactor" : kFactorValues(kFactor_13TeV=1.21),
+            "CrossSection" : XSValues(
+                XS_2016=0.03209, XSSource_2016="XSDB (LO)",
+                XS_2017=0.008001, XSSource_2017="XSDB (LO)",
```
kpedro88, line 399: this is a 4x change from 2016 -> 2017

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
-            "CrossSection" : XSValues(XS_13TeV=68.54),
-        },
-        "WJetsToQQ_HT-800toInf_qc19_3j" : {
-            "CrossSection" : XSValues(XS_13TeV=34.68),
-        },
-        "WJetsToQQ_HT-600ToInf" : {
-            "CrossSection" : XSValues(XS_13TeV=95.14),
-            "kFactor" : kFactorValues(kFactor_13TeV=1.21),
+            "CrossSection" : XSValues(
+                XS_13TeV=61526.7, XSSource_13TeV="https://twiki.cern.ch/twiki/bin/viewauth/CMS/SummaryTable1G25ns#W_jets NNLO (60430.0 @ NLO)",
```
kpedro88, line 414: is there a reason to keep this one? should a k-factor be assessed?

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+                XS_2016=50260.0, XSSource_2016="XSDB (LO)",
+                XS_2017=52940.0, XSSource_2017="XSDB (LO)",
+                XS_2018=52850.0, XSSource_2018="XSDB (LO)",
+            ),
+        },
+        "QCD_FlatPt_15_3000" : {
+            "CrossSection" : XSValues(
+                XS_2017=1985000000.0, XSSource_2017="XSDB (LO)",
+                XS_2018=1370000000.0, XSSource_2018="XSDB (LO)",
```
kpedro88: also surprised at this change 2017->2018

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+            "CrossSection" : XSValues(
+                XS_2016=1.613, XSSource_2016="XSDB",
+                XS_2017=1.079, XSSource_2017="GenXSecAnalyzer",
+                XS_2018=1.079, XSSource_2018="GenXSecAnalyzer",
+            ),
+        },
+        "QCD_Pt-15to7000_Flat" : {
+            "CrossSection" : XSValues(
+                XS_2016=1976000000.0, XSSource_2016="XSDB",
+                XS_2017=1370000000.0, XSSource_2017="XSDB",
```
kpedro88, line 579: i'm a bit suspicious that this is the exact value for QCD_FlatPt_15_3000 in 2018, can you double check

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+            ),
+        },
+        "QCD_Pt-15to7000_FlatP6" : {
+            "CrossSection" : XSValues(
+                XS_2016=1973000000.0, XSSource_2016="XSDB",
+            ),
+        },
+        "QCD_Pt-15to7000_Flat2017" : {
+            "CrossSection" : XSValues(
+                XS_2017=1361000000.0, XSSource_2017="XSDB",
```
kpedro88, line 590: why is this different from QCD_Pt-15to7000_Flat

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+                XS_2016=2762530.0, XSSource_2016="AN2017_013_v17, XSDB 2735000",
+                XS_2017=2336000.0, XSSource_2017="XSDB",
+                XS_2018=2343000.0, XSSource_2018="GenXSecAnalyzer",
+            ),
         },
         "QCD_Pt_120to170" : {
-            "CrossSection" : XSValues(XS_13TeV=471100.0),
+            "CrossSection" : XSValues(
+                XS_2016=471100.0, XSSource_2016="AN2017_013_v17, XSDB 467500",
+                XS_2017=407300, XSSource_2017="XSDB",
```
kpedro88: add .0 to these?

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+            "CrossSection" : XSValues(
+                XS_13TeV=6077.22, XSSource_13TeV="XSDB (NNLO)",
+            ),
         },
         "ZJetsToNuNu_HT-100To200" : {
-            "CrossSection" : XSValues(XS_13TeV=344.8305),
-            "kFactor" : kFactorValues(kFactor_13TeV=1.23),
+            "CrossSection" : XSValues(
+                XS_2016=93.35, XSSource_2016="XSDB (LO)",
+                XS_2017=302.8, XSSource_2017="XSDB (LO)",
```
kpedro88, line 809: really a 3x jump 2016 -> 2017?

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+                kFactor_2017=1.1374, kFactorSource_2017="XSDB DYJetsToLL_M-50 NNLO/LO=6077.22/5343",
+                kFactor_2018=1.1421, kFactorSource_2018="XSDB DYJetsToLL_M-50 NNLO/LO=6077.22/5321",
+            ),
         },
         "ZJetsToNuNu_HT-200To400" : {
-            "CrossSection" : XSValues(XS_13TeV=95.5341),
-            "kFactor" : kFactorValues(kFactor_13TeV=1.23),
+            "CrossSection" : XSValues(
+                XS_2016=25.85, XSSource_2016="XSDB (LO)",
+                XS_2017=92.59, XSSource_2017="XSDB (LO)",
```
kpedro88, line 825: these numbers all seem crazy

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+                kFactor_2017=1.1374, kFactorSource_2017="XSDB DYJetsToLL_M-50 NNLO/LO=6077.22/5343",
+                kFactor_2018=1.1421, kFactorSource_2018="XSDB DYJetsToLL_M-50 NNLO/LO=6077.22/5321",
+            ),
         },
         "ZJetsToNuNu_HT-400To600" : {
-            "CrossSection" : XSValues(XS_13TeV=13.1979),
-            "kFactor" : kFactorValues(kFactor_13TeV=1.23),
+            "CrossSection" : XSValues(
+                XS_2016=3.584, XSSource_2016="XSDB (LO)",
+                XS_2017=13.18, XSSource_2017="XSDB (LO)",
```
kpedro88, line 841: this is more than 4x

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
         "ZJetsToNuNu_Zpt-100to200" : {
-            "CrossSection" : XSValues(XS_13TeV=43.5543),
+            "CrossSection" : XSValues(
+                XS_2016=35.99, XSSource_2016="XSDB (LO)",
+            ),
         },
         "ZJetsToNuNu_Zpt-200toInf" : {
-            "CrossSection" : XSValues(XS_13TeV=5.28654),
-        },
-        "ST_s-channel_4f_hadronicDecays" : {
```
kpedro88, line 397: we seem to have lost this sample

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+            ),
         },
         "ST_t-channel_antitop_4f_inclusiveDecays" : {
-            "CrossSection" : XSValues(XS_13TeV=80.95),
+            "CrossSection" : XSValues(
+                XS_13TeV=80.95, XSSource_13TeV="Unknown",
+                XS_2017=67.91,  XSSource_2017="XSDB (NLO)",
+            ),
         },
         "ST_t-channel_antitop_4f_InclusiveDecays" : {
```
kpedro88: we should figure out a better way to avoid duplicating all the xsec values just because of bad GEN naming conventions

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+                XS_13TeV=19.4674, XSSource_13TeV="Unknown",
+                XS_2016=38.09,    XSSource_2016="XSDB (NLO)",
+                XS_2017=34.91,    XSSource_2017="XSDB (NLO)",
+            ),
         },
         "ST_tW_antitop_5f_inclusiveDecays" : {
-            "CrossSection" : XSValues(XS_13TeV=35.6),
+            "CrossSection" : XSValues(
+                XS_13TeV=35.6, XSSource_13TeV="Unknown",
+                XS_2016=38.06, XSSource_2016="XSDB (NLO)",
```
kpedro88, line 1055: i don't believe that inclusiveDecays can have the same xsec as NoFullyHadronicDecays
one of these values is wrong

kpedro88: consider comment at https://github.com/TreeMaker/TreeMaker/pull/505/files#diff-cc5b25d546afd2a07d4cc2600dcf2cedR830

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+            "CrossSection" : XSValues(
+                XS_13TeV=35.6, XSSource_13TeV="Unknown",
+                XS_2016=38.09,    XSSource_2016="XSDB (NLO)",
+                XS_2017=34.91,    XSSource_2017="XSDB (NLO)",
+            ),
         },
         "tZq_W_lept_Z_hadron_4f_ckm" : {
-            "CrossSection" : XSValues(XS_13TeV=0.0758),
-        },
-        "tZq_Zhad_Wlept_4f_ckm" : {
```
kpedro88, line 434: lost sample

WeightProducer/python/MCSampleValues.py
```diff
@@ -63,493 +63,1039 @@ class MCSampleValuesHelper():
...
+                XS_2016=0.6578, XSSource_2016="XSDB (LO)",
+                XS_2017=0.7158, XSSource_2017="GenXSecAnalyzer",
+                XS_2018=0.7153, XSSource_2018="GenXSecAnalyzer",
+            ),
         },
         "WWTo1L1Nu2Q" : {
-            "CrossSection" : XSValues(XS_13TeV=49.997),
+            "CrossSection" : XSValues(
+                XS_2016=45.68, XSSource_2016="XSDB (LO)",
+                XS_2017=80.74, XSSource_2017="XSDB (LO)",
```
kpedro88, line 1107: almost x2 jump?

WeightProducer/python/getWeightProducer_cff.py
```diff
@@ -56,7 +56,7 @@ def getWeightProducer(fileName, scan=False):
     for sample in samples:
         if sample.name in fileName and sample.production in fileName:
             weightProducer.Method     = cms.string(sample.Method)
-            weightProducer.XS         = cms.double(sample.XS)
+            weightProducer.XS         = cms.double(sample.XS*sample.kFactor)
```
kpedro88: should we consider storing k-factor in a separate branch also?

WeightProducer/python/MCSampleValues.py
aperloff: We don't understand this one yet. I can tell you that the Madgraph proc definitions are different, but I'm not really sure what effect the differences would have, if any. For reference, here are the relevant parts of the 2017 and 2018 process/madevent/Cards/proc_card_mg5.dat files:

2017:
```
import model sm-ckm_no_b_mass
define p = g u c d s u~ c~ d~ s~
define j = g u c d s u~ c~ d~ s~
define l+ = e+ mu+
define l- = e- mu-
define vl = ve vm vt
define vl~ = ve~ vm~ vt~
define p = 21 2 4 1 3 -2 -4 -1 -3 5 -5 # pass to 5 flavors
define j = p
define p = p b b~
define j = j b b~
define part =  ve vm vt e- mu- ve~ vm~ vt~ e+ mu+ ta- ta+
generate p p > t t~ a, t > b part part, t~ > b~ part part  @1
add process p p > t t~, t > b part part a, t~ > b~ part part  @2
add process p p > t t~, t > b part part , t~ > b~ part part a  @3
```

2018:
```
import model sm-no_b_mass
define p = g u c d s u~ c~ d~ s~
define j = g u c d s u~ c~ d~ s~
define l+ = e+ mu+
define l- = e- mu-
define p = 21 2 4 1 3 -2 -4 -1 -3 5 -5 # pass to 5 flavors
define j = p
define p = g u c d s b u~ c~ d~ s~ b~
define j = g u c d s b u~ c~ d~ s~ b~
define l+ = e+ mu+ ta+
define l- = e- mu- ta-
define vl = ve vm vt
define vl~ = ve~ vm~ vt~
generate p p > t t~ > l+ vl l- vl~ b  b~ a
```

WeightProducer/python/MCSampleValues.py
aperloff: I'll run the 2016 sample through the GenXSecAnalyzer and see if it matches the value there or if it's closer to the 2017/2018 values.

WeightProducer/python/MCSampleValues.py
aperloff: I used these values for assessing the k-factors used in the HT binned samples above. I don't think it hurts to have this value there given that the code should default to the 13TeV value and only go to the year-by-year values if it doesn't exist. I'm double checking that now.

WeightProducer/python/MCSampleValues.py
aperloff: This has more to do with the change in how the samples are flattened from year to year. I'm not so surprised by this.

WeightProducer/python/MCSampleValues.py
aperloff: Depends on if you'd like to use the "Flat" or "Flat2017" sample as a reference:
https://cms-gen-dev.cern.ch/xsdb/?columns=38472128&currentPage=0&ordDirection=1&ordFieldName=accuracy&pageSize=0&searchQuery=process_name%3DQCD%20%26%26%20process_name%3DFlat%20%26%26%20process_name%3D7000%20%26%26%20process_name%3DCP5

aperloff: I created a different section for the Flat2017 sample below, so I'd like to keep this value here.

WeightProducer/python/MCSampleValues.py
aperloff: Two different ways of scaling the cross section to flatten the pthat distribution.

WeightProducer/python/MCSampleValues.py
kpedro88: I'm not sure what you mean. It seems suspicious to me to get the same values for 15to3000 and 15to7000.

WeightProducer/python/MCSampleValues.py
aperloff: This has to do with the 2016 sample being generated using only a single lepton flavor and the 2017/2018 samples using all three lepton flavors. The 2016 cross section values will be multiplied by 3. For reference, I've put the Madgraph process fragments below:

2016:
```
generate p p > ve ve~ / h @0
add process p p > ve ve~ j / h @1
add process p p > ve ve~ j j / h @2
add process p p > ve ve~ j j j / h @3
add process p p > ve ve~ j j j j / h @4
```

2017/2018:
```
generate    p p > vl vl~ / h @0
add process p p > vl vl~ j / h @1
add process p p > vl vl~ j j / h @2
add process p p > vl vl~ j j j / h @3
add process p p > vl vl~ j j j j / h @4
```

WeightProducer/python/MCSampleValues.py
aperloff: See previous comment about ZJetsToNuNu.

WeightProducer/python/MCSampleValues.py
aperloff: See previous comment about ZJetsToNuNu.

WeightProducer/python/MCSampleValues.py
aperloff: I'm putting it back in. It's only available in 2017 and 2018. I have to run the GenXSecAnalyzer over this one.

WeightProducer/python/MCSampleValues.py
aperloff: Suggestions?

kpedro88: add an "alternate name" field?

WeightProducer/python/MCSampleValues.py
aperloff: Returned

WeightProducer/python/getWeightProducer_cff.py
aperloff: I think you mean as a separate member of the weightProducer and yes, I think that would be prudent. I was going to make the change in a different PR, but I can do it in this one if you'd like. Just let me know.

WeightProducer/python/MCSampleValues.py
kpedro88: or perhaps even easier to implement:
just define the alternate names as copies of the existing entries at the bottom of the file, after the dictionary is defined

WeightProducer/python/getWeightProducer_cff.py
kpedro88: I meant a separate branch in the output tree, but how it's handled in WeightProducer could be modified similarly. It's fine to make another PR for that (in that case, I would leave this one as it currently is, so the meaning of the "cross section" output value is preserved)

WeightProducer/python/MCSampleValues.py
aperloff: I've run the GenXSecAnalyzer for the 2016 sample and it confirms the cross section found in the XSDB. I'm still not sure why this cross section changed so dramatically. I'll ask some GEN contacts.

WeightProducer/python/getWeightProducer_cff.py
aperloff: Okay, I'll leave this alone for now. I'll make a different PR which takes the k-factor out of the XS value and puts it in a separate branch in the tree.

WeightProducer/python/MCSampleValues.py
aperloff: Should be replaced now.

WeightProducer/python/MCSampleValues.py
aperloff: We've found the issue here. The following samples have different phase space cuts (notice the "\_" vs "-"):
```
/TTGamma_Dilept_TuneCP5_13TeV-madgraph-pythia8/RunIIAutumn18MiniAOD-102X_upgrade2018_realistic_v15-v1/MINIAODSIM
/TTGamma_Dilept_TuneCP5_13TeV_madgraph_pythia8/RunIIAutumn18MiniAOD-102X_upgrade2018_realistic_v15-v1/MINIAODSIM
```
The sample using the "-" has a cross section of 1.495, while the the one with the "\_" has a cross section of 0.5799. TreeMaker uses the samples with "\_" for all three years, so this is the value taken.

WeightProducer/python/MCSampleValues.py
aperloff: Implemented.

WeightProducer/python/MCSampleValues.py
aperloff: I removed the following entries:
```
        "QCD_FlatPt_15_3000" : {
            "CrossSection" : XSValues(
                XS_2017=1985000000.0, XSSource_2017="XSDB (LO)",
                XS_2018=1370000000.0, XSSource_2018="XSDB (LO)",
            ),
        },
        "QCD_FlatPt_15_3000HS" : {
            "CrossSection" : XSValues(
                XS_13TeV=1356000000.0, XSSource_13TeV="Unknown",
            ),
        },
```
Not only could we not trust their values, but TreeMaker doesn't even contain file lists for these samples. I'd rather these just not pollute the code. These are names for relval samples anyway.

aperloff: Now that we only have 'QCD_Pt-15to7000*' samples I've double checked the values with the GenXSecAnalyzer. They should all make sense now.

WeightProducer/python/MCSampleValues.py
aperloff: In other words you want me to multiply the XS_2016 and XS_2017 values by the formula in that comment for the NoFullyHadronicDecays samples? Where does the 0.108 factor come from? Or do you just want me to confirm all of these values using the GenXSecAnalyzer?

kpedro88: 0.108 = Br(W->ℓν)

Can you run GenXsecAnalyzer and check for consistency with the expectation from branching fraction?

aperloff: I don't know of an option like that for the GenXSecAnalyzer. I re-ran it for all of the single top samples to confirm what is in the XSDB. They matched.

I'm going to run jobs to get the 2018 values as I realize they are missing. What would you like me to do about the NoFullyHadronicDecays samples?

kpedro88: I just meant compare the GenXsecAnalyzer values for Inclusive and NoFullyHadronicDecays. If they come out the same, then I guess we have to multiply the BR by hand.

aperloff: They came out the same for 2016 and 2017. I'm running 2018 now, but I don't expect there to be anything different.

Just to confirm, you want me to scale the NoFullyHadronicDecays like `XS*(1-(1-3*0.108)^2)`? I can then make a note of this in the XS_Source section so that we don't have to go through this exercise again.

WeightProducer/python/MCSampleValues.py
aperloff: This jump in XS is likely due to a mistake during production where, contrary to the sample name, the WWTo1L1Nu2Q sample does have events where both Ws decay leptonically (see https://twiki.cern.ch/twiki/bin/viewauth/CMS/MCKnownIssues#WWTo1L1Nu2Q_13_TeV_amcatnloFXFX). Given this, I propose to leave the cross section alone.

WeightProducer/python/MCSampleValues.py
aperloff: This seems to be related to the second issue (HT stitching) reported at https://twiki.cern.ch/twiki/bin/viewauth/CMS/MCKnownIssues#WJetsToLNu_HT_and_DYJets_HT_LO_M. There they recommend applying a set of "per-bin cross section correction factors ... derived based on comparing the HT distribution between the HT-binned and inclusive samples." I recommend we add another namedtuple_with_defaults type (i.e. correction_factors), where the default value is 1. Then we can scale the cross sections as recommended. This will maintain the provenance of the corrections and allow for maximum flexibility.

kpedro88: I think this is a good idea

WeightProducer/python/MCSampleValues.py
kpedro88: Yes

WeightProducer/python/MCSampleValues.py
aperloff: This should be implemented now.

WeightProducer/python/MCSampleValues.py
aperloff: Let me know what you think of this implementation.

aperloff:
I know the helper returns the correct values. I'm now double checking that the WeightProducer returns the correct values.

aperloff:
I ran three unit tests, two each with the Summer16v3.TTJets_SingleLeptFromT and 2018.WJetsToLNu_HT-600To800 samples. For each pair of tests I ran one before the WeightProducer modifications and one after the modifications. The before and after tests produced ntuples with the same weights up to the expected changes with the k-factors, branching ratios, and cross sections. I also calculated, by hand, the expected before and after weights, which confirm the values obtained by TreeMaker.

WeightProducer/python/MCSampleValues.py
```diff
@@ -50,506 +68,1240 @@ def get_year(self, mcVersion):
...
+            ),
         },
         "ST_tW_top_5f_NoFullyHadronicDecays" : {
-            "CrossSection" : XSValues(XS_13TeV=19.4674),
+            "CrossSection" : XSValues(
+                XS_2016=38.09, XSSource_2016="XSDB (NLO)",
+                XS_2017=34.91, XSSource_2017="XSDB (NLO)",
+                XS_2018=34.91, XSSource_2018="GenXSecAnalyzer",
+            ),
+            "BranchingRatio" : BRValues(
```
kpedro88, line 1032: make this a single BR_13TeV (also below)

WeightProducer/python/MCSampleValues.py
```diff
@@ -50,506 +68,1240 @@ def get_year(self, mcVersion):
...
-            "CrossSection" : XSValues(XS_13TeV=0.2118),
-        },
-        "ttHToNonbb_M125" : {
-            "CrossSection" : XSValues(XS_13TeV=0.2118),
+            "CrossSection" : XSValues(
+                XS_2016=0.4870, XSSource_2016="GenXSecAnalyzer",
+                XS_2017=0.5086, XSSource_2017="GenXSecAnalyzer",
+            ),
+            "BranchingRatio" : BRValues(
+                BR_2016=(1-0.577), BRSource_2016="https://twiki.cern.ch/twiki/bin/viewauth/CMS/SummaryTable1G25ns#ttH",
```
kpedro88: also BR_13TeV here

WeightProducer/python/MCSampleValues.py
aperloff: Changed

WeightProducer/python/MCSampleValues.py
aperloff: Changed

PR #516, opened 2020-05-21
[Run2_2017] Last minute changes needed for V18

aperloff:
This PR implements several changes needed for V18.
1. Add the SMS-TChiHH_HToBB_HToBB_2D file lists and dictionary entries.
2. Split the ttbar submission dictionary into two, TTJets* samples in one and TTTo* samples in another.
3. Fix get_py.py, which broke due to a change in the CRAB3 infrastructure.
4. Change the workaround needed for when jobs submitted from CMSConnect land on T1_US_FNAL. The user is set to "cmsgli" rather than the username of the person submitting the jobs. Since this prevents the job from xrdcp'ing the output to EOS. The workaround is to use gfal-copy. Previously we used the predefined CMSSITE and USER environment variables, but these are no longer being set. We now get the CMSSITE value from the job classads and use gfal-copy for any job which lands on T1_US_FNAL.


Production/python/get_py.py
```diff
@@ -2,7 +2,7 @@
 # check os
 slc = os.environ['SCRAM_ARCH'].split('_')[0]
 extra_paths = [
-    "/cvmfs/cms.cern.ch/crab3/"+slc+"_amd64_gcc493/cms/dbs3-client/3.5.1/lib/python2.7/site-packages/",
+    "/cvmfs/cms.cern.ch/share/cms/crab-prod/3.3.2005-bcolbf/lib/",
     "/cvmfs/cms.cern.ch/crab3/"+slc+"_amd64_gcc493/cms/dbs3-pycurl-client/3.5.1/lib/python2.7/site-packages/",
```
kpedro88: this should probably just be removed

aperloff: The pycurl path has been removed. I tested `get_py.py` and it still works.

PR #520, opened 2020-05-27
[Run2_2017] Update WGJets_MonoPhoton and ZJetsToNuNu_HT-400To600 samples (new versions)

aperloff:
Update the WGJets_MonoPhoton_Pt-G-40to130 sample because it was replaced with a newer version. This PR replaces the contents of the file list and replaces the WeightProducer line.


aperloff:
The second commit also updates the ZJetsToNuNu_HT-400To600 samples from v1 to v2.

PR #525, opened 2020-06-26
Preparing for EMJ Ntuple production

yimuchen:
1. Added missing dark quark PDGID
2. Removed a bad file in one sample
3. Added dictionary for EMJ sample for job submission


kpedro88:
@yimuchen please update https://github.com/TreeMaker/TreeMaker/blob/Run2_2017/WeightProducer/python/EMJsamples.py for the sample whose file you removed, otherwise the number of events for the normalization will be wrong.

yimuchen:
The EMJsamples.py file was created before that particular sample had finished production, so I had put in the default 5000 events by hand. I have made sure that the number of events are correct when compared with the samples in the file lists:  https://github.com/TreeMaker/TreeMaker/tree/Run2_2017/Production/python/PrivateSamples

Production/test/condorSub/dict_EMJ2016.py
```diff
@@ -0,0 +1,161 @@
+flist = {
+    "scenario": "Summer16v3sig",
+    "samples": [
+        ["PrivateSample.EMJ_2016_mMed-1000_mDark-10_ctau-0p1_unflavored-down"],
```
kpedro88: PrivateSample. -> PrivateSamples.
(everywhere)

yimuchen: Ah... that's a silly mistake

Utils/src/CandidateTrackFilter.cc
```diff
@@ -297,12 +297,14 @@ bool CandidateTrackFilter::filter(edm::StreamID, edm::Event& iEvent, const edm::
 	//-------------------------------------------------------------------------------------------------
 	loopOverCollection(infos,pfCandidates,ttBuilder,primaryVertex,true);
```
aperloff: You should add an isValid to the lostTracks and pfCandidates as well. Also, you should print out a warning if these collections are missing. Use the edm::LogWarning messageLogger. What do you thin, make sure this doesn't get printed every event, but only the first event where the collection is missing?

kpedro88: right now i haven't worried about printing messages every event or not

yimuchen: If we don't mind additional variables in the class, I guess printing the error message just for the first event should be fine. Let me see.

Utils/src/CandidateTrackFilter.cc
```diff
@@ -295,14 +295,34 @@ bool CandidateTrackFilter::filter(edm::StreamID, edm::Event& iEvent, const edm::
 	//-------------------------------------------------------------------------------------------------
 	// loop over the various track containing collections
 	//-------------------------------------------------------------------------------------------------
-	loopOverCollection(infos,pfCandidates,ttBuilder,primaryVertex,true);
-	loopOverCollection(infos,lostTracks,ttBuilder,primaryVertex,false);
-	loopOverCollection(infos,lostEleTracks,ttBuilder,primaryVertex,false);
+	static bool printErrorpfCandidates_  = false;
```
kpedro88: -1
statics are not thread safe
just print it every time

PR #528, opened 2020-07-17
[Run2_2017] Turn on branch splitting and replace TLorentzVector

aperloff:
This PR has two main components:
1. It allows us to turn on branch splitting as an option. This makes it easier for uproot/coffea to understand the branch structure. It's more numpy like. This was simple enough for most branches except for those involving `TLorentzVector`, thus...
2. Globally replace `TLorentzVector` with `ROOT::Math::PtEtaPhiEVector` (a.k.a `math::PtEtaPhiELorentzVector`. TLorentzVector is old, slow, and deprecated. The primary problem here is that it involves a custom streamer, which means it cannot be split. Moving to the more modern Lorentz vector classes allows for faster code, branch splitting, and a reduction in ntuple size.

Speaking of ntuple size, this PR also reduces the event size by ~29.3%. I produced identical ntuples, each of 100 events. The first one included `TLorentzVector` and a branch splitting of 0. That ntuple had an event size of 28.20 KiB. The second ntuple used the `math::PtEtaPhiELorentzVector` class and had a splitting of 99. That ntuple had an event size of 19.95 KiB.

When reviewing this PR, please pay particular attention to any place we use a constructor or one of the `SetCoordinates`/ `SetPxPyPzE` member functions. I tried to be very careful when modifying those parts of the code, but it would be good to have a second set of eyes look over it.


aperloff:
The previous tests were done in an V18e (EMJ Analysis) style configuration using the unit test command:
```bash
python unitTest.py test=2 command="emerging=True deepAK8=False deepDoubleB=False doZinv=False nestedVectors=False splitLevel=99" run=True fork=False
```

I re-ran the test using a V18 (SUSY) style unit test command:
```bash
python unitTest.py test=2 command="splitLevel=99" run=True fork=False
```
The tests showed even greater space saving potential, with a ~32% decrease in the compressed space per event (5.08 KiB before the change and 3.45 KiB after). I think the small 2.7% greater space savings might be due to the increase in the relative number of LorentzVectors used by the AK8, subjet, doubleB tag, and clean jet branches. All of the branches just mentioned would be replaced by 3 std::vector<ROOT::Math::LorentzVector> branches in the V18e ntuples, so it's logical that the space saving potential would be higher in the V18 ntuples.

aperloff:
I ran another test checking the values of the branches with `std::vector<ROOT::Math::PtEtaPhiEVector>` compared to those with `std::vector<TLorentzVector>`. I tested 100 events from a `Summer16v3.TTJets_SingleLeptFromT_TuneCUETP8M1_13TeV-madgraphMLM-pythia8` file with the V18e style ntuples. I was looking for differences in the size of the vectors and in the values stored in the vectors. There were no differences found, which is good.

Utils/src/GenLeptonRecoCand.cc
```diff
@@ -344,7 +340,7 @@ GenLeptonRecoCand::produce(edm::StreamID, edm::Event& iEvent, const edm::EventSe
       //      printf("No hadronic tracks!\n");
       selectedTauLeadTrkIso->push_back(-999.);
       selectedTauLeadTrkAct->push_back(-999.);
-      selectedTauLeadTrk->push_back(TLorentzVector(0,0,0,0));
+      selectedTauLeadTrk->push_back(math::PtEtaPhiELorentzVector(0,0,0,0));
```
kpedro88: this would be better as `emplace_back(0,0,0,0)`

Utils/src/Mt2Producer.cc
```diff
@@ -49,9 +49,9 @@ class Mt2Producer : public edm::global::EDProducer<> {
   edm::EDGetTokenT<reco::CandidateView> JetTok_;
   edm::EDGetTokenT<edm::View<pat::MET>> MetTok_;
 
-  vector<TLorentzVector> getHemispheres(vector<TLorentzVector> jets) const;
-  double getMT2Hemi(vector<TLorentzVector> jets, TLorentzVector metVec) const;
-  double computeMT2(TLorentzVector visaVec, TLorentzVector visbVec, TLorentzVector metVec) const;
+  vector<math::PtEtaPhiELorentzVector> getHemispheres(vector<math::PtEtaPhiELorentzVector> jets) const;
```
kpedro88: this function is never actually defined or used, so just remove it

Utils/src/Mt2Producer.cc
```diff
@@ -49,9 +49,9 @@ class Mt2Producer : public edm::global::EDProducer<> {
...
+  double getMT2Hemi(vector<math::PtEtaPhiELorentzVector> jets, math::PtEtaPhiELorentzVector metVec) const;
```
kpedro88: pass everything in this function and the next function as const ref
(should actually make this code a bit faster)

Utils/src/GenLeptonRecoCand.cc
aperloff: Done

Utils/src/Mt2Producer.cc
aperloff: Done

Utils/src/Mt2Producer.cc
aperloff: Done

aperloff:
All of the review comments have been addressed.

Issue #535, opened 2020-08-05
GenParticle charge branch for V18e


aperloff:
Currently the GenParticle collection does not store the charge of the particle. While this information is technically encoded in the PdgId, having every piece of code store and filter the neutral from charged PDGIDs is onerous and error prone. Also, trying to figure out which composite particles are charged vs neutral could be a pain. It should require very little space to include a new branch of ints. I propose to add this branch to the tree.

aperloff:
This issue was resolved by #537.

PR #537, opened 2020-08-06
[Run2_2017] Add charge branch to GenParticles

aperloff:
This PR adds a branch containing charge information to the GenParticles collection. This change will satisfy #535 .

### DIFF
--- a/Utils/src/GenParticlesProducer.cc
+++ b/Utils/src/GenParticlesProducer.cc
@@ -35,20 +35,24 @@ public:
 private:
     void produce(edm::StreamID, edm::Event&, const edm::EventSetup&) const override;
     const reco::GenParticle* findLast(const reco::GenParticle& particle) const;
-    void saveChain(int, int, int, const reco::GenParticle&,
+    void saveChain(int depth, int parentId, int parent_idx, const reco::GenParticle& particle,
                    std::unique_ptr<std::vector<math::PtEtaPhiELorentzVector>>& genParticle_vec,
-                   std::unique_ptr<std::vector<int>>&, std::unique_ptr<std::vector<int>>&,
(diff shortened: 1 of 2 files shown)


aperloff:
I forgot to say that this only changes the size of the ntuple by 0.15%. Using the TTJets_SingleLeptFromT sample and the settings `splitLevel=99 emerging=True deepAK8=False deepDoubleB=False doZinv=False nestedVectors=False` this amounts to a 0.03 KiB/event change.

Utils/src/GenParticlesProducer.cc
```diff
@@ -35,16 +35,19 @@ class GenParticlesProducer : public edm::global::EDProducer<> {
 private:
     void produce(edm::StreamID, edm::Event&, const edm::EventSetup&) const override;
     const reco::GenParticle* findLast(const reco::GenParticle& particle) const;
-    void saveChain(int depth, int parentId, int parent_idx, const reco::GenParticle& particle, std::unique_ptr<std::vector<math::PtEtaPhiELorentzVector>>& genParticle_vec,
-                   std::unique_ptr<std::vector<int>>& PdgId_vec, std::unique_ptr<std::vector<int>>& Status_vec,
-                   std::unique_ptr<std::vector<int>>& Parent_vec, std::unique_ptr<std::vector<int>>& ParentId_vec,
-                   std::unordered_set<const reco::Candidate *>& stored_particles_ref,
-                   std::vector<const reco::Candidate *>& stored_particles_list, std::vector<const reco::Candidate *>& parents_list) const;
+    void saveChain(int, int, int, const reco::GenParticle&,
```
kpedro88: prefer to keep variable names in member function definitions so the interface is more readable/understandable

aperloff: Ugh! You're one of those! ;) Okay, I've added the names back in.

PR #540, opened 2020-08-11
[Run2_2017] Store the index of the vertex associated to each track

aperloff:
Currently we store tracks and vertices in the TreeMaker ntuples. The reco::Tracks store the associated reco::VertexRef, but we don't look at that or store it. In the EMJ meeting on 08/10/2020 the idea of associating a track to a given PV came up. This PR adds a new std::vector<int> which lists the stored index for the vertex associated to each track in the ntuple. For the PV we didn't store the vector will contain an index of -1.

After running a unit test before an after the change the ntuple increased in size/event by ~0.7%.


aperloff:
We did some checks and this change doesn't seem to have an impact on the time taken by the PrimaryVerticesProducer or the Candidate TrackFilter.

PR #541, opened 2020-08-14
[Run2_2017] Add an EMJ related branch and track sorting

aperloff:
This PR does two things:
1. It adds a branch which contains the sum of the squares of the pT of the tracks associated to a given vertex. This increases the size of the ntuple by ~1% (tested using the EMJ settings on a Summer16v3 TTJets sample).
2. It sorts the tracks and their associated vectors by the track pT (decreasing pT)

### DIFF
--- a/Utils/src/CandidateTrackFilter.cc
+++ b/Utils/src/CandidateTrackFilter.cc
@@ -67,6 +67,21 @@ public:
 		pfcands_pvassociationquality = std::make_unique<vector<int>>();
 		pfcands_dzassociatedpv       = std::make_unique<vector<double>>();
 		pfcands_vtxidx               = std::make_unique<vector<int>>();
+		vtx_sumtrackpt2              = std::make_unique<vector<double>>();
+	}
+
+	void calculateVertexQuantities(const int & nvtx) {
+		for (auto ivtx=0; ivtx<nvtx; ivtx++) {
+			auto it = pfcands_vtxidx->begin();
+			int index = -1;
+			double sumpt2 = 0;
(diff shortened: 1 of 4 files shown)


Utils/src/CandidateTrackFilter.cc
```diff
@@ -67,6 +70,21 @@ class TrackInfos {
 		pfcands_pvassociationquality = std::make_unique<vector<int>>();
 		pfcands_dzassociatedpv       = std::make_unique<vector<double>>();
 		pfcands_vtxidx               = std::make_unique<vector<int>>();
+		vtx_sumtrackpt2              = std::make_unique<vector<double>>();
+	}
+
+	void calculateVertexQuantities(const int & nvtx) {
```
kpedro88: wouldn't it be more efficient to use the stored vertex indices that are already computed for the tracks? just loop over all tracks, adding the track pt2 to the vector entry corresponding to the vertex index

aperloff: Okay, I changed this function to use the existing indices. I checked and it's functionally equivalent.

kpedro88: primitive types should be passed by copy, rather than const ref

aperloff: Changed. Out of curiosity, why is that? I know it won't make functionally change anything, but is there a penalty in time or memory to passing a const ref?

kpedro88: Yes, it's actually faster to copy a primitive type than to pass a reference. For a complex class, it's faster to pass a reference.

PR #542, opened 2020-08-19
[Run2_2017] Store GenVertices for EMJ analysis

aperloff:
Store the GenVertices for the selected GenParticles and store an index to each vertex. This change has little to no effect on the size of the ntuple when tested using the EMJ settings on a Summer16v3 TTJets sample.


kpedro88:
+1

PR #543, opened 2020-08-19
[Run2_2017] Remove unnecessary offset branches

aperloff:
This change will introduce "associated" vector of vector types. Just like the regular ```vector<vector<X>>``` branches these can be saved in a flattened manner. However, the associated offsets branch will not be saved. It is assumed that there will be at least one branch with the same shape where the offsets are saved. For example, using this method 6 offsets branches can be removed from the `JetsAK8_subject_X` branches.

This change was tested using a Summer16v3 TTJets sample using the EMJ settings. In my tests it didn't seem to make much of a space difference weather or not the extra offsets branches were there. It could be because the number of subjets was really low. Still the lack of redundant information will lead to greater clarity, which is enough of a reason to go ahead with this change.


kpedro88:
+1

PR #544, opened 2020-08-19
Option to save counts rather offsets when saving nested vectors in a flat way

tklijnsma:
Implements a command line option "nestedCounts"; if set to True, counts will be saved rather than offsets when saving nested arrays in a flattened way (i.e. using nestedVectors=False)

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,8 @@ Brief explanation of the options in [maker.py](./TreeMaker/python/maker.py)
 * `saveMinimalGenParticles`: save only the hard scatter gen particles coming from top decays, boson decays, semi-visible jets, or SUSY particles (default=True)
 * `saveGenTops`: save the 4-vectors of the generated tops and the TTbar reweighting scale factor (default=False)
 * `doMT2`: switch to enable the storage of the MT2 variable (default=False)
-* `nestedVectors`: switch to change from saving vector<\vector<T>> to saving vector<T> values and vector<\int> offsets (default=True)
+* `nestedVectors`: switch to change from saving vector<\vector<T>> to saving vector<T> values and vector<\int> counts (default=True)
+* `nestedCounts`: if set to False, stores offsets rather than counts when using `nestedVectors=False` (default=True)
 
 The following parameters take their default values from the specified scenario:
 * `globaltag`: global tag for CMSSW database conditions (ref. [FrontierConditions](https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideFrontierConditions))
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ Brief explanation of the options in [maker.py](./TreeMaker/python/maker.py)
 * `saveGenTops`: save the 4-vectors of the generated tops and the TTbar reweighting scale factor (default=False)
 * `doMT2`: switch to enable the storage of the MT2 variable (default=False)
 * `nestedVectors`: switch to change from saving vector<\vector<T>> to saving vector<T> values and vector<\int> counts (default=True)
-* `nestedCounts`: if set to False, stores offsets rather than counts when using `nestedVectors=False` (default=True)
+* `storeOffsets`: if set to True, stores offsets rather than counts when using `nestedVectors=False` (default=False)
 
 The following parameters take their default values from the specified scenario:
 * `globaltag`: global tag for CMSSW database conditions (ref. [FrontierConditions](https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideFrontierConditions))
--- a/TreeMaker/interface/TreeMaker.h
+++ b/TreeMaker/interface/TreeMaker.h
@@ -66,7 +66,7 @@ class TreeMaker : public edm::one::EDAnalyzer<edm::one::SharedResources> {
 		edm::Service<TFileService> fs;
 		string treeName;
 		TTree* tree;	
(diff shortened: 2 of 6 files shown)


TreeMaker/interface/TreeMaker.h
```diff
@@ -376,24 +376,24 @@ class TreeNestedVector : public TreeObject<std::vector<std::vector<Base>>> {
...
+			TreeObject<Top>(tempFull_,title_,splitLevel_), nestedVectors(nestedVectors_), nestedCounts(nestedCounts_), associated(associated_) {}
 		// Destructor
 		~TreeNestedVector() override {}
 		
 		// Functions
 		// From: https://stackoverflow.com/questions/17294629/merging-flattening-sub-vectors-into-a-single-vector-c-converting-2d-to-1d
 		void flatten(Top const& all, Sub &accum, vector<int> &offsets) {
 			// Don't store any offsets if there are no sub-vectors
-			if (!associated && all.size() > 0)	offsets.insert(std::end(offsets),0);
+			if (!associated && all.size() > 0){ if(!nestedCounts) offsets.insert(std::end(offsets),0); }
```
kpedro88: simplify this to one if condition: `if (!associated && all.size() > 0 && !nestedCounts)`

TreeMaker/python/maker.py
```diff
@@ -158,7 +159,10 @@ def printSetup(self):
         print " Storing a minimal set of GenParticles: "+str(self.saveMinimalGenParticles)
         print " Storing the GenTops: "+str(self.saveGenTops)
         print " Saving the MT2 variable: "+str(self.doMT2)
-        if self.nestedVectors: print " Saving nested vectors as vector<vector<T>>"
+        if self.nestedVectors:
+            print " Saving nested vectors as vector<vector<T>>"
+            if self.nestedCounts:
+                print " Saving counts(not offsets) for nested vectors"
```
kpedro88: space after counts

kpedro88:
In addition, just to confirm:
1. Was the output with this PR checked to make sure the branches had the expected content?
2. Is this terminology and schema consistent with coffea etc.? (the word "counts", omitting the first 0)

aperloff:
Do we want Offsets to be the default and not counts? If the goal of this PR is to make TreeMaker compatible with Coffea/awkward1/uproot4, shouldn't counts be the default?

TreeMaker/interface/TreeMaker.h
tklijnsma: for `nestedCounts` the `0` shouldn't be inserted, but the `else return` should be skipped, hence this ugly double `if`. The alternative would be an `else if (nestedCounts) ; else ...`?

kpedro88:
The PR has it enabled by default in maker.py and treeMaker.py...

but that reminds me, please add this parameter to the readme.

TreeMaker/interface/TreeMaker.h
kpedro88: ah, then it's fine as is.

tklijnsma:
@aperloff  I think I did make counts the default (i.e. nestedCounts=True)

@kpedro88  I did produce two files and checked whether the output looks as expected. The beginning `0` is only needed for offsets (offsets is like slices, and the first slice always starts at 0; counts is just plainly the length of the nested array).

TreeMaker/python/maker.py
tklijnsma: Done

tklijnsma:
Updated the readme

README.md
```diff
@@ -253,7 +253,8 @@ Brief explanation of the options in [maker.py](./TreeMaker/python/maker.py)
 * `saveMinimalGenParticles`: save only the hard scatter gen particles coming from top decays, boson decays, semi-visible jets, or SUSY particles (default=True)
 * `saveGenTops`: save the 4-vectors of the generated tops and the TTbar reweighting scale factor (default=False)
 * `doMT2`: switch to enable the storage of the MT2 variable (default=False)
-* `nestedVectors`: switch to change from saving vector<\vector<T>> to saving vector<T> values and vector<\int> offsets (default=True)
+* `nestedVectors`: switch to change from saving vector<\vector<T>> to saving vector<T> values and vector<\int> counts (default=True)
+* `nestedCounts`: if set to False, stores offsets rather than counts when using `nestedVectors=False` (default=True)
```
kpedro88: I think you mean "when using `nestedVectors=True`"?

tklijnsma: so `nestedVectors=True` means not doing the flattening, `nestedVectors=false` means do the flattening. If doing the flattening, `nestedCounts=False` would use offsets rather than counts.

I would almost suggest changing both cl options to `flattenVectors` and `flattenCounts`...

kpedro88: Okay, this is definitely getting confusing.

By itself, it makes sense if `nestedVectors=True` means store `vector<vector<T>>`. But it's very confusing if `nestedCounts` only applies when `nestedVectors` is disabled. I think we need to change `nestedCounts` to something clearer. Maybe `storeOffsets`, set to `False` by default, so it's clear that the default option is now storing counts?

We may also want to disable `nestedVectors` by default at this point.

tklijnsma:
Okay I flipped the flag to `storeOffsets` rather than `nestedCounts` (reversed definition). Maybe let me test it once more before merging it in

aperloff:
I'm happy to say this PR is now passing the integration tests.

+1

PR #558, opened 2020-11-17
Added additional gen particles for EMJ analysis

yimuchen:
Saving the decay chain from the HV scalar mediator (pdgid:4900001)


TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -157,15 +157,15 @@ def makeTreeFromMiniAOD(self,process):
...
                 # Particles we want to save the last copy from the hard scatter
                 parentIds = cms.vint32(
                     6,22,23,24,25,
                     1000021,1000022,1000023,1000024,1000025,1000035,1000037,1000039,
                     1000001,1000002,1000003,1000004,1000005,1000006,
                     2000001,2000002,2000003,2000004,2000005,2000006,
-                    4900021,4900023,4900101,4900102,4900103,4900111,4900113,4900211,4900213,51,52,53,
-                    5000001,5000002,
+                    4900001,4900021,4900023,4900101,4900102,4900103,4900111,4900113,4900211,4900213,51,52,53,
+                    5000001,5000002,#
```
aperloff: Why is there a `#` at the end of the line?

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -157,15 +157,15 @@ def makeTreeFromMiniAOD(self,process):
                 genCollection = cms.InputTag("prunedGenParticles"),
                 debug = cms.bool(False),
                 # Particles we want to save from the decay chain of the tops
-                childIds = cms.vint32(1,2,3,4,5,11,12,13,14,15,16,24),
+                childIds = cms.vint32(1,2,3,4,5,6,11,12,13,14,15,16,24),
```
aperloff: In the past we have not added the top quark ID as one of the child IDs to save because the top is not a final state particle. Is there a reason why you are adding it here? Have you checked what affect this will have on TTJets samples and signal samples? My guess is that it will have little to no effect on the signal samples, but will start to save a bunch of extra particles for TTJets or T1tttt/T2tt samples.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -157,15 +157,15 @@ def makeTreeFromMiniAOD(self,process):
...
                 # Particles we want to save the last copy from the hard scatter
                 parentIds = cms.vint32(
                     6,22,23,24,25,
                     1000021,1000022,1000023,1000024,1000025,1000035,1000037,1000039,
                     1000001,1000002,1000003,1000004,1000005,1000006,
                     2000001,2000002,2000003,2000004,2000005,2000006,
-                    4900021,4900023,4900101,4900102,4900103,4900111,4900113,4900211,4900213,51,52,53,
-                    5000001,5000002,
+                    4900001,4900021,4900023,4900101,4900102,4900103,4900111,4900113,4900211,4900213,51,52,53,
```
kpedro88: Does it make sense to add all the scalar mediators here? (also 4900002, etc.)

yimuchen: I guess it depends on how future proof we want the TreeMaker code to be? Right now, we are setting 4900001 to be the only mediator that can be probed by LHC, do we expect anyone to use the other scale mediators?

TreeMaker/python/makeTreeFromMiniAOD_cff.py
yimuchen: That was mainly due to concerns about the top type coupling of the flavored model. I preemptively added this without thinking too much about what this would do to other samples, but after testing with our private samples, I don't think we need this change. I will be removing this in the next update

TreeMaker/python/makeTreeFromMiniAOD_cff.py
kpedro88: I think the MadGraph t-channel setup can produce the other mediators.

yimuchen: OK, I guess this change doesn't taking up more space in the output anyway, for the sake of completion we might as well add them now.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
yimuchen: Solved. I actually don't know how this got here...

Utils/src/GenParticlesProducer.cc
```diff
@@ -219,6 +219,7 @@ enum particle_type
     sbottom2=2000005,
     stop2=2000006,
     //HV particles
+    hvscalar=4900001,
```
aperloff, line 222: Should we also preemptively add particle_type enums for multiple scalars (4900001-4900006)?

yimuchen: Right now, I don't thing we need to. These enums are currently only used for flagging the start and end of a PDGID range to be stored (line 377-378 of the same file), so unless futures users add algorithms that required more sophisticated decay topology crawling, having just 4900001 to mark the staring point should be enough.

aperloff:
+1

PR #568, opened 2021-04-03
Added flag for adding photon information

yimuchen:
When running the TreeMaker, one can add the option `doPhotons=True` to force the addition of photon variables. 
The corresponding variables is also added using the same routine with the `doZinv=True` flag


kpedro88:
thanks @yimuchen ! @lwang046 before we merge this, can you check that the photon ID and isolation working points we use by default are adequate for what you want to do?

yimuchen:
All the mechanicse are there, with the difference being that the 2018 paper uses the medium working points, while the default cut values in TreeMaker uses the loose working point. We are not actually using the photons for any computation, just as an offline signal rejection target, so maybe changing working points doesn't matter much?

kpedro88:
Right; we can always apply a tighter working point offline, as long as all the ingredients are there.

lwang046:
It seems indeed all the ingredients needed according to this [twiki](https://twiki.cern.ch/twiki/bin/view/CMS/CutBasedPhotonIdentificationRun2) are there.

yimuchen:
Pull request submitted: https://github.com/TreeMaker/TreeMaker/pull/569

PR #578, opened 2021-08-11
[Run2_2017] Save gen jet truth information needed to tag emerging jets

aperloff:
This PR will add an option to save the number of gen jet constituents originating from a hidden valley ancestor. This is then saved, along with the total multiplicity, for both AK4 and AK8 gen jets when the `emerging` flag is set. This change adds 0.08 KiB to each event (without=29.91 KiB/evt, with= 29.99 KiB/evt).

The PR also adds two unit tests which are useful for testing the Fall17 era for EMJ.


aperloff:
Here is a plot showing the fraction of gen jet constituents with an HV ancestor.
[Screen Shot 2021-08-11 at 4.03.08 PM.pdf](https://github.com/TreeMaker/TreeMaker/files/6971394/Screen.Shot.2021-08-11.at.4.03.08.PM.pdf)

PR #579, opened 2021-09-01
Added SVJ tchannel jet categorization and gen-level truth MT2 algorithms

Keane-Tan:


kpedro88:
Additional note: this PR also includes an option to get low-pt AK8 jets by reclustering. Right now, all of the jet substructure variables are recomputed, which can be quite slow (especially for ECFs). In the future, we may implement some approach to reuse the existing jet substructure variables from the default AK8 jet collection, and only recompute them for the lower-pt jets, but this is nontrivial to implement. (Right now, only one optimization is included: DeepAK8 is not rerun for low-pt AK8 jets.)

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -1072,15 +1117,25 @@ def makeTreeFromMiniAOD(self,process):
...
-            DarkQuarkID = cms.uint32(4900101),
-            DarkMediatorID = cms.uint32(4900023),
+            GenJetTag = GenJetAK8Tag,
+            coneSize = cms.double(0.8),
+            DarkStableIDs = cms.vuint32(51,52,53),
+            DarkQuarkIDs = cms.vuint32(4900101),
+            DarkMediatorIDs = cms.vuint32(4900023),
+            DarkHadronIDs = cms.vuint32(4900111,4900113,4900211,4900213),
+            DarkGluonIDs = cms.vuint32(4900021),
+            SMQuarkIDs = cms.vuint32(1,2,3,4,5,6,7,8),
```
kpedro88: 7,8 are not SM quark PDG IDs.

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -1072,15 +1117,25 @@ def makeTreeFromMiniAOD(self,process):
...
         )
+        if self.tchannel:
+            process.HiddenSector.DarkQuarkIDs = [4900101,4900102]
+            process.HiddenSector.DarkMediatorIDs = [4900001,4900002,4900003,4900004,4900005,4900006]
         self.VarsDouble.extend([
             'HiddenSector:MJJ(MJJ_AK8)',
             'HiddenSector:Mmc(Mmc_AK8)',
             'HiddenSector:MT(MT_AK8)',
+            'HiddenSector:tchMT2(tchMT2_AK8)',
```
kpedro88: I would name this branch `GenMT2_AK8`

kpedro88:
Also: this PR needs https://github.com/TreeMaker/JetToolbox/pull/2 to function correctly.

Utils/src/HiddenSectorProducer.cc
```diff
@@ -22,143 +27,390 @@
...
+    else{
+      if(isDark(DarkQuarkIDs_,dau)){
+        firstQdM.insert(dau);
+        if(secondDM == false){
+          firstQdM1 = dau;
+          secondDM = true;
+        }
+        else firstQdM2 = dau;
+      }
+      else if(isDark(SMQuarkIDs_,dau)){
```
kpedro88: now that we're using `isDark` with SM quarks also, maybe we should change the name to `isParticle` (otherwise it's confusing)

Utils/src/HiddenSectorProducer.cc
```diff
@@ -22,143 +27,390 @@
...
+          firstQsM1 = dau;
+          secondSM = true;
+        }
+        else firstQsM2 = dau;
+      }
+    }
+  }
+}
+
+void HiddenSectorProducer::firstDark(CandPtr part, CandSet& firstMd, CandSet& firstQd, CandSet& firstGd, CandSet& firstQdM, CandSet& firstQsM, CandPtr& firstQdM1, CandPtr& firstQdM2, CandPtr& firstQsM1, CandPtr& firstQsM2, bool& secondDM, bool& secondSM) const {
```
kpedro88, line 124: It's hard to follow what this function does when the actual `firstDark` sets are mixed together with the objects needed for `medDecay`. Can you reorganize it so all the `firstDark` arguments come first, and then the `medDecay`-only arguments come second? and add some more comments to indicate everything that's happening

Utils/src/HiddenSectorProducer.cc
```diff
@@ -22,143 +27,390 @@
...
+          medDecay(dau,firstQdM,firstQsM,firstQdM1,firstQdM2,firstQsM1,firstQsM2,secondDM,secondSM);
+        }
+        else if (isDark(DarkQuarkIDs_,dau)) firstQd.insert(dau);
+        else if (isDark(DarkGluonIDs_,dau)) firstGd.insert(dau);
+      }
+    }
+  }
+}
+
+int HiddenSectorProducer::checkLast(const reco::GenJet& jet, CandSet stableDs, int value, double& frac) const {
```
kpedro88: second arg should be `const CandSet& stableDs`

Utils/src/HiddenSectorProducer.cc
```diff
@@ -22,143 +27,390 @@
...
+int HiddenSectorProducer::checkFirst(const reco::GenJet& jet, const CandSet& firstP, int value) const {
+  //compare first particles using jet cone (can't compare constituents' parents because all dark hadrons descend from all first dark particles, strong force problems)
+  for(const auto& part : firstP){
+    if(reco::deltaR(jet, *part) < coneSize_) return value;
+  }
+  return 0;
+}
+
+// match particles from mediator to jets
+void HiddenSectorProducer::matchPFMtoJet(CandPtr& part, std::vector<LorentzVector>& matchedJets, const reco::GenJet& jet, int& nPartPerJet, int& t_MT2JetID, const int& mt2ID) const {
```
kpedro88: `part`, `mt2ID` don't need to be passed as references

Utils/src/HiddenSectorProducer.cc
```diff
@@ -22,143 +27,390 @@
...
+  if(reco::deltaR(jet,*part) < coneSize_ and jet.pt() > 100)
```
kpedro88: remind me about the reason for the jet pt cut?

Utils/src/HiddenSectorProducer.cc
```diff
@@ -22,143 +27,390 @@
...
+  {
+    matchedJets.push_back(jet.p4());
```
kpedro88: avoid copying four-vectors unnecessarily: just use a vector of `reco::GenJet*` (then access `.p4()` when needed)

Utils/src/HiddenSectorProducer.cc
```diff
@@ -22,143 +27,390 @@
...
+  fillSet(DarkQuarkIDs_,"DarkQuarkIDs",iConfig);
+  fillSet(DarkHadronIDs_,"DarkHadronIDs",iConfig);
+  fillSet(DarkGluonIDs_,"DarkGluonIDs",iConfig);
+  fillSet(DarkStableIDs_,"DarkStableIDs",iConfig);
+  fillSet(SMQuarkIDs_,"SMQuarkIDs",iConfig);
+  //combined list of possible first particles
+  DarkFirstIDs_.insert(DarkMediatorIDs_.begin(),DarkMediatorIDs_.end());
+  DarkFirstIDs_.insert(DarkQuarkIDs_.begin(),DarkQuarkIDs_.end());
+  DarkFirstIDs_.insert(DarkGluonIDs_.begin(),DarkGluonIDs_.end());
+
```
kpedro88, line 261: add a call to `asymm_mt2_lester_bisect::disableCopyrightMessage();` here to block the unnecessary printout

Utils/src/HiddenSectorProducer.cc
```diff
@@ -22,143 +27,390 @@
...
+          }
+        }
+      }
+      jet_isHV_vec->push_back(matched);
+    }
+
+    //t-channel jet categorization
+    //gen particle organization: last daughters of last copies of dark hadrons, first copies of dark quarks/gluons/mediators, first copies of quarks/gluons/SM quarks from mediator
+    //naming scheme: dark particles Pd, SM particles Ps; P = D (generic daughters), Q (quarks), G (gluons), M (mediators)
+    CandPtr firstQdM1, firstQdM2, firstQsM1, firstQsM2;
```
kpedro88: as noted above, try to organize `firstDark` objects vs `medDecay` objects separately to make things clearer

Utils/src/HiddenSectorProducer.cc
```diff
@@ -22,143 +27,390 @@
...
+    CandSet stableDs, firstMd, firstQd, firstGd, firstQdM, firstQsM;
+    std::vector<LorentzVector> dQM1Js,dQM2Js,SMM1Js,SMM2Js;
```
kpedro88: as noted above, avoid copying four-vectors unnecessarily

Utils/src/HiddenSectorProducer.cc
Keane-Tan: In a pair production event, we expect 2 mediators, where each mediator decays to 1 dark quark and 1 SM quark. So we expect 2 dark quarks and 2 SM quarks from the mediators in such an event.

When determining which jets to use to compute the MT2, I decided to impose the following conditions:
1. Each jet can only contain exactly one dark/SM quark from the mediators.
2. Each dark/SM quark from the mediators can only fall into the radius of exactly one jet.

Otherwise, there is no way to tell how the energy from the quark is split among the jets that are sharing the same quark or how to split the energy from the jet if more than one quark falls inside the jet. Therefore events that do not fulfill the conditions above are simply vetoed when it comes to computing the MT2.

If there is no pT requirement for the jet, we see a lot of overlap between jets. This overlap increases the likelihood of violating the conditions, so we end up with less events for computing the MT2.

Keane-Tan: Maybe this is only true for the signal (3000_20_0.3_peak) I have been testing. I guess for a much lower mMed, the pT requirement might become too high for the jet...

kpedro88: Yes, I understand what you mean about having multiple jets matching the mediator daughters. How well does it work if you just keep the highest-pT matching jet, ignoring any others? Or, alternatively, just use all the matching jets (adding up their four-vectors)...

kpedro88: I guess the most fundamental problem is if daughters from two different mediators are too close together. Maybe those are the cases we should exclude.

Keane-Tan: I think I will remove the pT cut for now.

Utils/src/HiddenSectorProducer.cc
```diff
@@ -22,143 +27,436 @@
...
+    void medDecay(CandPtr part, CandSet& firstQdM, CandSet& firstQsM, CandPtr& firstQdM1, CandPtr& firstQdM2, CandPtr& firstQsM1, CandPtr& firstQsM2, bool& secondDM, bool& secondSM) const;
+    void firstDark(CandPtr part, CandSet& firstMd, CandSet& firstQd, CandSet& firstGd, CandSet& firstQdM, CandSet& firstQsM, CandPtr& firstQdM1, CandPtr& firstQdM2, CandPtr& firstQsM1, CandPtr& firstQsM2, bool& secondDM, bool& secondSM) const;
+    int checkLast(const reco::GenJet& jet, const CandSet& stableDs, int value, double& frac) const;
+    int checkFirst(const reco::GenJet& jet, const CandSet& firstP, int value) const;
+    void matchPFMtoJet(CandPtr part, std::vector<const reco::GenJet*>& matchedJets, const reco::GenJet& jet, int& nPartPerJet, int& t_MT2JetID, int mt2ID) const;
+    double calculateMT2(edm::Handle<edm::View<reco::GenMET>> h_genmets, const reco::GenJet*& dQM1J, const reco::GenJet*& SMM1J, const reco::GenJet*& dQM2J, const reco::GenJet*& SMM2J) const;
+    double calculateMT2(edm::Handle<edm::View<reco::GenMET>> h_genmets, const LorentzVector& dQM1J, const LorentzVector& SMM1J, const LorentzVector& dQM2J, const LorentzVector& SMM2J) const;
```
kpedro88: This version of the function seems unused, so it could be removed

Utils/src/HiddenSectorProducer.cc
```diff
@@ -22,143 +27,436 @@
...
+  return asymm_mt2_lester_bisect::get_mT2(
+    FJet0.M(), FJet0.Px(), FJet0.Py(),
+    FJet1.M(), FJet1.Px(), FJet1.Py(),
+    METx, METy, 0.0, 0.0, 0
+  );
+}
+
+const reco::GenJet* HiddenSectorProducer::highPt(std::vector<const reco::GenJet*>& Js) const {
+  double largestPt = 0;
+  const reco::GenJet* largestPtJ;
```
kpedro88: initialize this to `nullptr` (there was a compiler warning: `warning: 'largestPtJ' may be used uninitialized in this function [-Wmaybe-uninitialized]`)

PR #580, opened 2021-09-01
add r/f scale uncertainty recalculation for pythia-only

kpedro88:
This PR adds the capability to recalculate renormalization and factorization scale uncertainties for MC samples generated with Pythia only.

The renormalization scale calculation relies on user input regarding the number of EM and QCD vertices in the physics process. Pythia functions are used directly to calculate the EM and strong couplings to ensure consistency.

Also, a bug is fixed: the 2017/18 PDF set was also used for 2016 by default, but 2016 MC uses a different PDF set.

### DIFF
--- a/Utils/src/PDFWeightProducer.cc
+++ b/Utils/src/PDFWeightProducer.cc
@@ -2,6 +2,8 @@
 #include <cmath>
 #include <memory>
 #include <mutex>
+#include <vector>
+#include <string>
 
 // user include files
 #include "FWCore/Framework/interface/Frameworkfwd.h"
@@ -12,11 +14,13 @@
 #include "FWCore/Framework/interface/MakerMacros.h"
 
 #include "FWCore/ParameterSet/interface/ParameterSet.h"
-
+#include "FWCore/Utilities/interface/Exception.h"
 #include "SimDataFormats/GeneratorProducts/interface/LHEEventProduct.h"
 #include "SimDataFormats/GeneratorProducts/interface/GenEventInfoProduct.h"
 
 #include "TVector2.h"
+
+#include <Pythia8/Pythia.h>
 
 enum class wtype { scale = 0, pdf = 1, ps = 2 };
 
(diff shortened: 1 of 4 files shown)


Utils/src/PDFWeightProducer.cc
```diff
@@ -201,6 +213,70 @@ void PDFWeightProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::Ev
...
+         LHAPDF::usePDFMember(1,0);
+         pdf1 = LHAPDF::xfx(1, x1, Q, id1)/x1;
+         pdf2 = LHAPDF::xfx(1, x2, Q, id2)/x2;
+         pdf1up = LHAPDF::xfx(1, x1, kUp*Q, id1)/x1;
+         pdf2up = LHAPDF::xfx(1, x2, kUp*Q, id2)/x2;
+         pdf1dn = LHAPDF::xfx(1, x1, kDn*Q, id1)/x1;
+         pdf2dn = LHAPDF::xfx(1, x2, kDn*Q, id2)/x2;
+      }
+      //for debugging
+      //std::cout << "pdf1 = " << pdf1 << ", pdf1up = " << pdf1up << ", pdf1dn = " << pdf1dn << ", pdf2 = " << pdf2 << ", pdf2up = " << pdf2up << ", pdf2dn = " << pdf2dn << std::endl;
```
aperloff: It would be better to add a `debug` flag and then use an `edm::MessageLogger` instance. That way you don't need to keep this comment around.

Utils/src/PDFWeightProducer.cc
```diff
@@ -201,6 +213,70 @@ void PDFWeightProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::Ev
...
+      auto coup = pythia_->couplingsPtr;
+      double Q2 = Q*Q;
+      double alpEM = coup->alphaEM(Q2);
+      double alpEMup = coup->alphaEM(kUp*kUp*Q2);
+      double alpEMdn = coup->alphaEM(kDn*kDn*Q2);
+      double alpS = coup->alphaS(Q2);
+      double alpSup = coup->alphaS(kUp*kUp*Q2);
+      double alpSdn = coup->alphaS(kDn*kDn*Q2);
+      //for debugging
+      //std::cout << "alpEM = " << alpEM << ", alpEMup = " << alpEMup << ", alpEMdn = " << alpEMdn << ", alpS = " << alpS << ", alpSup = " << alpSup << ", alpSdn = " << alpSdn << std::endl;
```
aperloff: Same comment about using a flag+message logger.

PR #587, opened 2021-10-07
[Run2_UL] Forward port changes from older branches

kpedro88:
This rebases the following PRs (which went into older branches after Run2_UL split off):
https://github.com/TreeMaker/TreeMaker/pull/568
https://github.com/TreeMaker/TreeMaker/pull/569
https://github.com/TreeMaker/TreeMaker/pull/574
https://github.com/TreeMaker/TreeMaker/pull/578
https://github.com/TreeMaker/TreeMaker/pull/579
https://github.com/TreeMaker/TreeMaker/pull/580
https://github.com/TreeMaker/TreeMaker/pull/581
https://github.com/TreeMaker/TreeMaker/pull/583
https://github.com/TreeMaker/TreeMaker/pull/589

Some "pointless" changes (addition of private pre-legacy samples, modifying pre-legacy unit tests, etc.) were omitted.

In addition https://github.com/TreeMaker/JetToolbox/pull/2 was included upstream as https://github.com/cms-jet/JetToolbox/pull/79, so I just updated https://github.com/TreeMaker/JetToolbox/tree/jetToolbox_102X to match the upstream.


Production/test/data/dict_xsec_Zprime.txt
```diff
@@ -1,47 +1,47 @@
...
-4300	0.0008757
-4400	0.0007037
-4500	0.0005655
-4600	0.0004544
-4700	0.0003652
-4800	0.0002934
-4900	0.0002358
-5000	0.0001895
-5100	0.0001523
+500	76.11
```
aperloff, line 1: Why did these XS change?

WeightProducer/python/SVJxsecs.py
```diff
@@ -145,5 +145,56 @@
     "mZprime-5100" : {
         "CrossSection" : XSValues(XS_13TeV=0.0001523),
     },
+    # Full t-channel cross section
+    "t-channel_mMed-200" : {
+        "CrossSection" : XSValues(XS_13TeV=990.8), # uncertainty: 3.617
+    },
+    # "t-channel_mMed-400" : {
```
aperloff, line 154: Why are these commented out?

Production/test/data/dict_xsec_Zprime.txt
kpedro88: They were wrong before: we obtained them from the dijet search, but they had a gen-level kinematic acceptance factor included. These are the actual cross section values.

WeightProducer/python/SVJxsecs.py
kpedro88: @Keane-Tan can you comment on this?

Keane-Tan: It was because the t-channel pair production has different cross section compared to the full t-channel production mode (t-channel + associated production + pair production). I used the same keys in this file when I produce the t-channel ntuples in both production modes, so I commented out the cross section values for the full t-channel production when I am making ntuples for the pair production only signals.

kpedro88: Okay, I've added this info in a comment. For the UL signal production, we should come up with a better naming scheme to avoid this ambiguity.

PR #590, opened 2021-10-12
[Run2_2017_V18] Add new FastSim samples

aperloff:
Add new FastSim files for SUS-20-004 and other similar SUSY analyses. This PR adds the new file lists and updates the relevant condorSub dictionaries.


Utils/python/signalscan_cfi.py
```diff
@@ -3,5 +3,8 @@
 SignalScanProducer = cms.EDProducer("SignalScanProducer",
 	signalType = cms.string("None"),
     debug = cms.bool(False),
-    isLHE = cms.bool(True)
+    isLHE = cms.bool(True),
+    genCollection = cms.InputTag("prunedGenParticles"),
+    motherPDGID = cms.int32(1000023),
+    lspPDGID = cms.int32(1000022),
```
kpedro88: I would move these settings into the `elif` above in `makeTree`, since they are only relevant for that case

Utils/src/SignalScanProducer.cc
```diff
@@ -8,6 +8,8 @@
 #include "FWCore/Utilities/interface/Exception.h"
 #include "FWCore/MessageLogger/interface/MessageLogger.h"
 #include "FWCore/ParameterSet/interface/ParameterSet.h"
+#include "DataFormats/Candidate/interface/Candidate.h"
+#include <DataFormats/HepMCCandidate/interface/GenParticle.h>
```
kpedro88: use quotes

Utils/src/SignalScanProducer.cc
```diff
@@ -186,20 +220,31 @@ void SignalScanProducer::getSUSYModelInfo(std::string comment){
 	//underscore-delimited data
 	parse::process(comment,'_',fields);
 
-	//several possible formats:
-	//model name_mMother_mLSP (1+2 fields)
-	//model name_xChi_mMother_mLSP (1+3 fields)
-	//model name_name_name_mMother_mLSP (3+2 fields)
-	//just take last two values and convert to doubles
+	if (type_ == signal_type::SUSYGenPart) {
```
kpedro88: personally, I would prefer this to be in its own function, rather than being shoved into the middle of a function that's handling another case

Utils/python/signalscan_cfi.py
aperloff: Done

Utils/src/SignalScanProducer.cc
aperloff: Done

Utils/src/SignalScanProducer.cc
aperloff: Okay, I've changed this.

Utils/src/SignalScanProducer.cc
```diff
@@ -202,6 +242,26 @@ void SignalScanProducer::getSUSYModelInfo(std::string comment){
 	signalParameters_.push_back(lspMass_);
 }
 
+//parse model comment for SUSYGenPart
+void SignalScanProducer::getSUSYGenPartModelInfo(std::string comment){
```
kpedro88: just make this take `const GenLumiInfoHeader& gen` to avoid messing with `getSUSYComment()`

Utils/src/SignalScanProducer.cc
```diff
@@ -133,7 +160,7 @@ SignalScanProducer::beginLuminosityBlock(edm::LuminosityBlock const& iLumi, edm:
 		if(debug_) edm::LogInfo("TreeMaker") << "SignalScanProducer: checking GenLumiInfoHeader";
 		edm::Handle<GenLumiInfoHeader> gen_header;
 		iLumi.getByToken(genLumiHeaderToken_, gen_header);
-		if(type_==signal_type::SUSY) getSUSYComment(*gen_header);
+		if(type_==signal_type::SUSY || type_==signal_type::SUSYGenPart) getSUSYComment(*gen_header);
```
kpedro88: make these two separate cases

Utils/src/SignalScanProducer.cc
aperloff: Changed

Utils/src/SignalScanProducer.cc
aperloff: Done

kpedro88:
+1

PR #592, opened 2021-10-22
Misc cleanup and fixes [UL]

kpedro88:
* Add some missing documentation
* Cleanup several producers
* Remove several unused producers
* Change defaults related to Coffea compatibility (increase splitlevel, disable nested vectors)
* Update unit tests
* Update 2017 JER factors
* Various JetToolbox-related fixes (goes with https://github.com/cms-jet/JetToolbox/pull/81; https://github.com/TreeMaker/JetToolbox already updated accordingly)
* Fix handling of `cms.bool` types
* Add 23% speedup in PUPPI
* reuse PUPPI weights when possible
* avoid unnecessary copies
* double->float conversion (enabled by default)
* Partial fix for displacedStandaloneMuons (emerging jets)

@aperloff the last item needs your attention: the type of the collection changed, and I'm not sure what information we need from it or how to access it.


TreeMaker/src/TreeMaker.cc
```diff
@@ -54,6 +54,7 @@ TreeMaker::TreeMaker(const edm::ParameterSet& iConfig) :
 	nestedVectors = iConfig.getParameter<bool>("nestedVectors");
 	storeOffsets = iConfig.getParameter<bool>("storeOffsets");
 	splitLevel = iConfig.getParameter<int>("splitLevel");
+    saveFloat = iConfig.getParameter<bool>("saveFloat");
```
aperloff: Spacing seems off here.

PR #598, opened 2021-12-03
[Run2_UL] Update the PDF Weight Code

aperloff:
Make sure the PDF and scale weights are recalculated when running over EMJ samples. Also update the PDFWeightProducer to correctly handle when the incoming partons are gluons. Rather than setting the pdgid to 21, as is the usual convention, the LHAPDF code wants the pdgid to be 0, as in [-5, -4, -3, -2, -1, 21, 1, 2, 3, 4, 5]. This is the same convention shown in the LHAPDF set for [NNPDF 3.1 NLO 0118](https://lhapdfsets.web.cern.ch/current/NNPDF31_nlo_as_0118/NNPDF31_nlo_as_0118.info). The idea for using 0 rather than 21 for the PDGID came from the [LHAPDF6 Design Guide](https://lhapdf.hepforge.org/design.html).

I did my best to confirm that the updated values are correct. I tried to make sure that the relative sizes of `newpdf1` and `newpdf2` were correct given `Q`, `x1`, `x2`, and Figure 3.1 (pg 23) from [arXiv:1706.00428](https://arxiv.org/pdf/1706.00428.pdf). As best I can tell, the values being returned are correct. 101 non-zero values are being returned for the PDF weights and 9 non-zero values are being returned for the scale weights.

This PR relies upon #597.

### DIFF
--- a/Utils/src/PDFWeightProducer.cc
+++ b/Utils/src/PDFWeightProducer.cc
@@ -192,9 +192,11 @@ void PDFWeightProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::Ev
 
       int id1 = genHandle->pdf()->id.first;
       double x1 = genHandle->pdf()->x.first;
+      id1 = std::abs(id1) == 21 ? 0 : id1;
 
       int id2 = genHandle->pdf()->id.second;
       double x2 = genHandle->pdf()->x.second;
+      id2 = std::abs(id2) == 21 ? 0 : id2;
 
       unsigned nweights = 1;
       if(LHAPDF::numberPDF(1)>1) nweights += LHAPDF::numberPDF(1);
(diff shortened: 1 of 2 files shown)


Utils/src/PDFWeightProducer.cc
```diff
@@ -192,9 +192,11 @@ void PDFWeightProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::Ev
 
       int id1 = genHandle->pdf()->id.first;
       double x1 = genHandle->pdf()->x.first;
+      id1 = std::abs(id1) == 21 ? 0 : id1;
```
kpedro88: can you make one function that does this, so the pattern doesn't have to be repeated everywhere?

Utils/src/PDFWeightProducer.cc
```diff
@@ -258,7 +271,9 @@ void PDFWeightProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::Ev
       double alpSup = coup->alphaS(kUp*kUp*Q2);
       double alpSdn = coup->alphaS(kDn*kDn*Q2);
       if(debug_)
-        edm::LogInfo("TreeMaker") << "PDFWeightProducer: alpEM = " << alpEM << ", alpEMup = " << alpEMup << ", alpEMdn = " << alpEMdn << ", alpS = " << alpS << ", alpSup = " << alpSup << ", alpSdn = " << alpSdn << std::endl;
+        edm::LogInfo("TreeMaker") << "PDFWeightProducer: alpEM = " << alpEM << ", alpEMup = " << alpEMup
+                                  << ", alpEMdn = " << alpEMdn << ", alpS = " << alpS << ", alpSup = " << alpSup\
```
kpedro88: remove errant \ ?

Utils/src/PDFWeightProducer.cc
aperloff: Done

Utils/src/PDFWeightProducer.cc
aperloff: Fixed

Issue #600, opened 2021-12-22
"ProductNotFound" When Running HiddenSectorProducer on UL MiniAOD


JHiltbrand:
In running `TreeMaker` with the `Run2_UL` branch, I find weird behavior when the `HiddenSectorProducer` is run. Seemingly pseudo-randomly, the producer encounters an InvalidID exception. An example exception message that occurred after successfully running on 36k events in the same ROOT file is shown below. Thus far, I have observed this to occur for 2016 and 2018 POWHEG ttbar files (only samples tested with).

```
----- Begin Fatal Exception 22-Dec-2021 15:31:49 UTC-----------------------
An exception of category 'ProductNotFound' occurred while
   [0] Processing  Event run: 1 lumi: 87331 event: 87330791 stream: 0
   [1] Running path 'WriteTree'
   [2] Prefetching for module TreeMaker/'TreeMaker2'
   [3] Calling method for module HiddenSectorProducer/'HiddenSector'
Exception Message:
InvalidID get by product ID: invalid ProductID supplied
   Additional Info:
      [a] If you wish to continue processing events after a ProductNotFound exception,
add "SkipEvent = cms.untracked.vstring('ProductNotFound')" to the "options" PSet in the configuration.

----- End Fatal Exception -------------------------------------------------
22-Dec-2021 15:31:49 UTC  Closed file root://cmsxrootd.fnal.gov//store/mc/RunIISummer20UL16MiniAODv2/TTToHadronic_TuneCP5_13TeV-powheg-pythia8/MINIAODSIM/106X_mcRun2_asymptotic_v17-v1/70000/0E17501A-465E-1D4F-970C-A44B4FAB86E0.root
```

kpedro88:
This is an unfortunately not very useful error message.

Does it really happen randomly, not on a specific event? Or does it just look that way because of e.g. running multithreaded?

If it's really random, then we need to run valgrind to check for memory corruption.

JHiltbrand:
I looked into one of the files that was being processed (different than example above) when the exception occurred, and `TreeMaker` will consistently encounter the same exception on the same event. For instance with this dilepton ttbar file [1]. The above exception is encountered on record 470.

[...]

PR #602, opened 2021-12-28
restrict gen-level hidden sector algorithms to SVJ signal samples

kpedro88:
This is both an optimization and a workaround for #600. The algorithms in `HiddenSectorProducer` that produce various gen-level jet IDs and related quantities for SVJ samples don't need to be run for SM samples (which won't contain any HV particles).

In addition, this PR removes some ambiguity in the module configuration. The `tchannel` flag is used to enable low-pT AK8 jet reclustering, which may be desired for any sample (signal, background, or data). However, this flag was also used to control some settings for the `HiddenSectorProducer` that were specific to t-channel SVJ signals, which is a different usage. The latter are now just controlled by the usual checks for an SVJ signal sample. Separate s- and t-channel mediator PDG ID lists are provided for the relevant algorithms so that everything can coexist.

attn: @jhiltbrand @Keane-Tan @cmadrid1


aperloff:
+1

PR #603, opened 2021-12-29
option to store all jet constituents efficiently

kpedro88:
This PR introduces a new producer that makes a single filtered collection of PF candidates, including only those candidates that are a constituent of some jet in a given list of jet collections. Each jet in each jet collection gets a list of indices into the filtered PF candidate collection. In this way, the constituents for jets from multiple collections can be stored without any duplication of information.

Open question: currently only the candidate 4-vector information is stored. Should other information also be stored, such as the PDG ID?

With this PR and other recent PRs, the updates I had scheduled for the first UL production are largely complete, so I have run some profiling studies that can be found here: https://www.dropbox.com/s/d9z6be0qghu3ruc/treemaker%20UL%20performance%20dec%2029%202021.pdf?dl=0

The production proposal should be discussed in more detail: @aperloff @jhiltbrand @cmadrid1 

It would also be good if someone could validate the output of this PR (i.e. ensuring that all the constituents are available and stored in a usable way).


kpedro88:
One possibility to save space: the jet constituents could be stored only for events passing some basic cut. A filter would produce a bool indicating if the selection passes, which the JetConstituents producer would consume and then only fill the vectors of constituent info if the bool was true. This does restrict the utility of the constituents in the ntuples to analyses with signal regions tighter than the basic cut, but that may be a necessary tradeoff.

kpedro88:
Another possibility: only store the jet constituents for certain jets passing some cut (e.g. on pT).

kpedro88:
The PDG ID is now added. Options/features that will be implemented in future PRs if/when needed:
* event-level filters
* jet-level filters
* additional constituent features

For now, the option added in this PR will be used to make mini-ntuples (w/ unneeded features, branches disabled, and only a small subset of samples) specifically for DNN training (rather than including constituents in all centrally-produced ntuples).

aperloff:
+1

PR #604, opened 2022-01-04
move to CMSSW_10_6_29_patch1

kpedro88:
The latest [10_6_X release](https://github.com/cms-sw/cmssw/releases/tag/CMSSW_10_6_29_patch1) includes my patch to speed up PUPPI. Moving to the new release means that we no longer have to merge and compile it in the local release area.


aperloff:
You also need to edit the following files:
- .gitlab-ci.yml
- .github/workflows/dockerimage.yml
- .docker/Dockerfile
- .docker/setupStandalone.sh

aperloff:
Well, it looks like the changes you made were successful. The code can compile and the tests run pseudo-successfully. However, because the `Summer16` scenario no longer exists the pre-merge integration test is no longer successful. My guess is that this has been true since we switched to the `Run2_UL` branch and removed the `Summer16` scenario. I really do need to make sure that there is a check on cmsRun which will cause the CI job to fail if that doesn't complete successfully.

aperloff:
The problem with the integration test not properly failing has been at least partially addressed in #606. While technically somewhat risky, I don't think that issue should hold up this PR.

PR #605, opened 2022-01-04
add option to specify extra args in condor dicts

kpedro88:
This option is added to help automate the application of special settings to groups of samples for ntuple production. An example is provided.


aperloff:
+1

PR #606, opened 2022-01-09
[Run2_UL] Non-zero unitTest return code on failed test

### LINKED ISSUE #508: PR integration test not triggering a job failure
When a PR triggers the GitHub actions workflow and integration test is performed after the build. This test runs a unit test. However, if the unit test fails, this does not constitute a failure of the action (see https://github.com/TreeMaker/TreeMaker/pull/505/checks?check_run_id=644339683). This should be fixed so that when unitTest.py fails, so does the job.

aperloff:
Have the unit test code exit with a non-zero return code if the cmsRun command doesn't complete successfully. This is needed because right now the integration tests in the CI are passing even when cmsRun doesn't complete successfully. The problem is that the non-zero cmsRun return code isn't being passed on by the python parent process. Since the parent process completes with exit code 0, the GitHub action believes that everything completed successfully. Now, the python process will exit with a non-zero return code when cmsRun doesn't finish with return code 0. This should make it so that the unit tests fail when cmsRun fails.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -53,10 +53,10 @@ jobs:
     - name: Run the container and perform an integration test
       if: github.event_name == 'pull_request'
       env:
-        download_url: https://cernbox.cern.ch/index.php/s/DsfTPyDEgCjTmBQ/download
-        file_name: eos.opendata.cms.MonteCarlo2016.RunIISummer16MiniAODv2.QCD_Pt_600to800_TuneCUETP8M1_13TeV_pythia8.MINIAODSIM.PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v6-v1.70000.0048131D-3CB3-E611-813A-001E67DFFB31_100evt.root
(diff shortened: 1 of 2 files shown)


aperloff:
As you can see with https://github.com/TreeMaker/TreeMaker/runs/4755508129?check_suite_focus=true, the integration test now appropriately fails.

The downside is that I don't see a Run2_UL compatible CERN Open Data dataset. The file we were using for the integration tests was a Summer16 MiniAODv2 dataset. Now that the Summer16 scenario and all of the 80X compatible code has been removed, there are no currently available open data files that meet our needs.

aperloff:
I have updated the file used for the integration tests and the GitHub Action so that the tests now complete successfully (https://github.com/TreeMaker/TreeMaker/runs/4769534260?check_suite_focus=true). This is a temporary solution until there is a more official "open data" file compatible with CMSSW_10_6_X (UL) available. However, for right now I think it's okay to merge this PR.

kpedro88:
+1

PR #607, opened 2022-01-12
disable zinv reclustering by default, but still keep basic photon variables

kpedro88:
I also removed explicit settings that are now defaults from the EMJ unit tests.


kpedro88:
Here are some updated profiles aimed toward the latest production plan: https://www.dropbox.com/s/ilbdds0yseg980e/treemaker%20UL%20performance%20jan%2012%202022.pdf?dl=0

PR #612, opened 2022-01-21
gen flavor for AK8 jets

kpedro88:
These quantities were requested for some DeepAK8-related studies. Adding them has a negligible impact on the file size.


aperloff:
+1

PR #617, opened 2022-02-08
Add feature to reduce the precision of single-precision floating point values

aperloff:
This PR adds a new feature to allow the user to set the number of bits used in the mantissa of single-precision floating point values. The idea is that if more of the bits in the mantissa are predictably set to zero, then the compression algorithm will be able to be more efficient.

I have tested the code and found no difference in total size or size/event before this PR or after this PR, but not changing the precision. When the precision was changed I found that the size/event decreased by ~3.5% (ever so slightly increasing as the number of events/file increased).

I haven't yet done:
- A scan over the precision to see the relationship between number of bits in the mantissa (i.e. space saved) and the loss of precision in any given branch.
- An analysis of the change in speed of the ntuplization process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,7 @@ Brief explanation of the options in [maker.py](./TreeMaker/python/maker.py)
 * `storeOffsets`: if set to True, stores offsets rather than counts when using `nestedVectors=False` (default=False)
 * `splitLevel`: split level for output TBranches (default=99)
 * `saveFloat`: convert doubles to floats in output (default=True)
+* `reduceFloatPrecision`: reduce the precision of the mantissa in the single precision floating point values in the output tree to the specified number of bits (default=-1 (off), minimum=0, maximum=23)
 
 The following parameters take their default values from the specified scenario:
 * `globaltag`: global tag for CMSSW database conditions (ref. [FrontierConditions](https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideFrontierConditions))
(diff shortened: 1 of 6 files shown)


aperloff:
A scan of the size of the ntuple (per event) versus the number of bits in the mantissa.
SizePerEntryVsNBits

aperloff:
**Timing test methodology:** The tests were performed on cmslpc113 with approximately 3 other people connected, none of them doing anything particularly compute intensive. There was plenty of memory and disk space available as well as nearly zero CPU usage. The tests were performed using [igProf](https://igprof.org/) through the [runIgprof.sh](https://github.com/kpedro88/utilities/blob/master/runIgprof.sh) script. I then looked at the total time taken as output in the igsorted*.res files. The input file was local and came from the dataset /TTJets_TuneCP5_13TeV-amcatnloFXFX-pythia8/RunIISummer20UL17MiniAODv2_106X_mc2017_realistic_v9-v2/MINIAODSIM. The test was run over 5000 events.

**Results of the timing tests:** Prior to the changes, the total time taken was 152.69 s (0.030538 s/evt -> 32.75 evt/s). After the changes, but keeping the original 23 bits in the mantissa, the time is 159.89 s (0.031978 s/evt -> 31.27 evt/s). After the changes and reducing the precision to 16 bits in the mantissa, the time rose negligibly from the initial time to 152.79 s (0.030558 s/evt -> 32.73 evt/s).

**Takeaway:** This PR results in a negligible timing change and is swamped by the change in time due to effects from using shared computing resources.

aperloff:
Same information in graphical form
TimePerEntry

PR #618, opened 2022-02-17
[Run2_UL] Job resubmission based on input file location

aperloff:
Add the ability to resubmit TreeMaker jobs based on the input file site location as reported by DAS. This funcationality currently relies on a new python module (file_finder_resubmitter.py). This commit adds the code changes for TreeMaker. Before this will work, a new PR will need to be opened to [kpedro88/CondorProduction](https://github.com/kpedro88/CondorProduction) in order to change a few things in [jobSubmitter.py](https://github.com/kpedro88/CondorProduction/blob/master/python/jobSubmitter.py).

New dependencies:
- This feature relies heavily on [manageJobs.py](https://github.com/kpedro88/CondorProduction/blob/master/python/manageJobs.py) from [kpedro88/CondorProduction](https://github.com/kpedro88/CondorProduction). It was decided not to put this functionality in there because the log parsing is TreeMaker specific.
  - Could we add a new ClassAd containing the specific input file to each job submitted by [jobSubmitter.py](https://github.com/kpedro88/CondorProduction/blob/master/python/jobSubmitter.py)? In that case we wouldn't need to parse logs and we could avoid being TreeMaker specific. Then this feature could be added add a new module to [kpedro88/CondorProduction](https://github.com/kpedro88/CondorProduction) and invoked from within [manageJobs.py](https://github.com/kpedro88/CondorProduction/blob/master/python/manageJobs.py).

Future updates:
- Eventually I would like tighter integration with jobSubmitterTM.py. This might entail moving all of the code from file_finder_resubmitter.py into jobSubmitterTM.py.
- Avoid parsing the log files. The jobSubmitterTM.py knows the number of each job and the inputFilesConfig. Therefore, one could imaging importing inputFilesConfig and getting the entry for that specific job.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ Python:
 * `-i, --ignore-args`: ignore args specified in the input dict (default = False)
 * `--offset [num]`: offset for arg file naming in chain jobs (default = 0)
 * `--maxJobs [num]`: max number of jobs to run (default = -1 -> not applied)
-* `-r, --resubmit`: resubmit the jobs based on where the input files are located (default = False)
+* `-r, --resubmit`: resubmit the jobs based on where the input files are located (new operation mode, default = False)
 * `--resubmit_options`: Options to be passed to file_finder_resubmitter.py, remember to add quotes around the string (default = "")
 
 Shell (in [step2.sh](./Production/test/condorSub/step2.sh)):
--- a/README.md
+++ b/README.md
@@ -120,6 +120,8 @@ Python:
 * `-i, --ignore-args`: ignore args specified in the input dict (default = False)
 * `--offset [num]`: offset for arg file naming in chain jobs (default = 0)
 * `--maxJobs [num]`: max number of jobs to run (default = -1 -> not applied)
+* `-r, --resubmit`: resubmit the jobs based on where the input files are located (default = False)
+* `--resubmit_options`: Options to be passed to file_finder_resubmitter.py, remember to add quotes around the string (default = "")
 
 Shell (in [step2.sh](./Production/test/condorSub/step2.sh)):
 * `-o [dir]`: output directory
--- a/Production/test/condorSub/file_finder_resubmitter.py
+++ b/Production/test/condorSub/file_finder_resubmitter.py
@@ -1,0 +1,130 @@
+import argparse
+import os
+import subprocess
+import sys
+if os.environ["CMSSW_BASE"] is not "":
+    try:
+        sys.path.append(os.environ["CMSSW_BASE"] + "/src/Condor/Production/python/")
+        import manageJobs
+    except ModuleNotFoundError as e:
+        print(e)
+else:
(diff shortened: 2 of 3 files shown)


kpedro88:
> Before this will work, a new PR will need to be opened to [kpedro88/CondorProduction](https://github.com/kpedro88/CondorProduction) in order to change a few things in [jobSubmitter.py](https://github.com/kpedro88/CondorProduction/blob/master/python/jobSubmitter.py).

What changes are needed in CondorProduction?

> Eventually I would like tighter integration with jobSubmitterTM.py

To me, the import looks fine, unless there's some functionality that is missed by doing it that way.

README.md
```diff
@@ -120,6 +120,8 @@ Python:
 * `-i, --ignore-args`: ignore args specified in the input dict (default = False)
 * `--offset [num]`: offset for arg file naming in chain jobs (default = 0)
 * `--maxJobs [num]`: max number of jobs to run (default = -1 -> not applied)
+* `-r, --resubmit`: resubmit the jobs based on where the input files are located (default = False)
```
kpedro88: I would note this explicitly as a new mode of operation (à la getpy in https://github.com/cms-svj/SVJProduction)

aperloff:
> What changes are needed in CondorProduction?

I took over the `-r` option for `--resubmit`, so I needed to change the shortened version of `--resub` (now called `--resub_script`) to `-R` in [jobSubmitter.py](https://github.com/kpedro88/CondorProduction/blob/master/python/jobSubmitter.py). I did this so that all of the job mode options has the same case.

> To me, the import looks fine, unless there's some functionality that is missed by doing it that way.

Nope, all the functionality of the new module is available. The thing that is missing or unavailable this way is the ability not to have to parse log files. That said, I'm still exploring other ways to avoid this.

kpedro88:
I would prefer not to introduce an API-breaking change in CondorProduction. I suggest renaming this new mode instead. Perhaps `-x, --xrootd-resubmit` or similar, noting the relationship to the similar `manageJobs` option?

aperloff:
`-x` is taken by `--redir`. I propose changing the shorter version of `--use-folders` to `-F` (was `-f`) and then use `-f` for `--resubmit` (as in resubmit by file).

kpedro88:
Ah, I forgot about the TreeMaker-specific arguments. Your proposal is fine with me.

README.md
aperloff: Done

aperloff:
This functionality has been implemented in https://github.com/kpedro88/CondorProduction/pull/14. I would prefer having the functionality in [manageJobs.py](https://github.com/kpedro88/CondorProduction/blob/master/python/manageJobs.py) anyway as it makes sense to have the resubmission-by-site infrastructure alongside the other resubmission utilities. In addition, that is more general location for this functionality, as opposed to being a TreeMaker only feature. Therefore, this PR is being deprecated.

PR #619, opened 2022-02-17
[Run2_UL] Add ClassAd containing input file(s) to HTCondor jobs

aperloff:
Insert an HTCondor ClassAd into the job based upon the files being processed by TreeMaker. This needs to take place once the input file(s) have been found. The earliest location for this, which only gets run once, is withing `maker.py`. There is a check to make sure this is only done if the job is running on a condor worker node and on an interactive node.

The only dependencies this adds to `maker.py` are on the python modules `os` and `subprocess`.

I don't foresee any slowdown or adverse effects coming from this as it only takes place once and the command being run is very fast.

### DIFF
--- a/TreeMaker/python/maker.py
+++ b/TreeMaker/python/maker.py
@@ -114,18 +114,21 @@ class maker:
                 value = "\"" + ",".join(self.readFiles) + "\""
                 try:
                     chirp_command = "condor_chirp set_job_attr_delayed " + key + " " + value
+                    if self.verbose:
+                        print chirp_command
                     proc = subprocess.Popen(chirp_command.split(), shell = False, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
                     outs, errs = proc.communicate()
                 except Exception as e:
                     proc.kill()
-                    print "An exception occurred: {}".format(e)
+                    print "An exception occurred when adding classad {}: {}".format(key,e)
                 finally:
                     if proc.returncode:
                         print "WARNING::condor_chirp failed to set the attribute \'" + key + "\'" 
                     else:
-                        print "HTCondor classad \'" + key + "\' set to " + value
+                        if self.verbose:
+                            print "HTCondor classad \'" + key + "\' set to " + value
         except KeyError:
-            print "<Not on a condor worker node>"
+            pass
 
         # branches for treemaker
         self.VectorRecoCand                 = cms.vstring()
--- a/TreeMaker/python/maker.py
+++ b/TreeMaker/python/maker.py
@@ -1,5 +1,8 @@
 import FWCore.ParameterSet.Config as cms
 
+import os
+import subprocess
+
 # import functions to be assigned as class methods
 from TreeMaker.TreeMaker.makeTreeFromMiniAOD_cff import makeTreeFromMiniAOD
 from TreeMaker.TreeMaker.JetDepot import JetVariations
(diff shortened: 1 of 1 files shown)


TreeMaker/python/maker.py
```diff
@@ -103,6 +106,27 @@ def __init__(self,parameters):
...
+                except Exception as e:
+                    proc.kill()
+                    print "An exception occurred: {}".format(e)
+                finally:
+                    if proc.returncode:
+                        print "WARNING::condor_chirp failed to set the attribute \'" + key + "\'" 
+                    else:
+                        print "HTCondor classad \'" + key + "\' set to " + value
+        except KeyError:
+            print "<Not on a condor worker node>"
```
kpedro88: I don't think this printout is necessary for typical use. would change to `pass` or similar if you don't want to catch non-`KeyError` exceptions

TreeMaker/python/maker.py
```diff
@@ -103,6 +106,27 @@ def __init__(self,parameters):
...
+            if os.environ["_CONDOR_CHIRP_CONFIG"]:
+                key = "ChirpTreeMakerReadFiles"
+                value = "\"" + ",".join(self.readFiles) + "\""
+                try:
+                    chirp_command = "condor_chirp set_job_attr_delayed " + key + " " + value
+                    proc = subprocess.Popen(chirp_command.split(), shell = False, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+                    outs, errs = proc.communicate()
+                except Exception as e:
+                    proc.kill()
+                    print "An exception occurred: {}".format(e)
```
kpedro88: I would make this message more precise e.g. `"An exception occurred when adding classad {}: {}".format(key,e)`

TreeMaker/python/maker.py
```diff
@@ -103,6 +106,27 @@ def __init__(self,parameters):
...
         self.readFiles = [(self.redir if val.startswith("/") else "")+val for val in self.readFiles]
 
+        # https://htcondor.readthedocs.io/en/latest/man-pages/condor_chirp.html?highlight=set_job_attr_delayed#chirp-commands
+        # https://github.com/cms-sw/cmssw/blob/master/FWCore/Services/plugins/CondorStatusUpdater.cc#L377-L410
+        try:
+            if os.environ["_CONDOR_CHIRP_CONFIG"]:
+                key = "ChirpTreeMakerReadFiles"
+                value = "\"" + ",".join(self.readFiles) + "\""
+                try:
+                    chirp_command = "condor_chirp set_job_attr_delayed " + key + " " + value
```
kpedro88: print the command if `verbose`?

TreeMaker/python/maker.py
```diff
@@ -103,6 +106,27 @@ def __init__(self,parameters):
...
+                    proc = subprocess.Popen(chirp_command.split(), shell = False, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+                    outs, errs = proc.communicate()
+                except Exception as e:
+                    proc.kill()
+                    print "An exception occurred: {}".format(e)
+                finally:
+                    if proc.returncode:
+                        print "WARNING::condor_chirp failed to set the attribute \'" + key + "\'" 
+                    else:
+                        print "HTCondor classad \'" + key + "\' set to " + value
```
kpedro88, line 130: this could also go behind `verbose`

TreeMaker/python/maker.py
```diff
@@ -103,6 +106,27 @@ def __init__(self,parameters):
...
+                try:
+                    chirp_command = "condor_chirp set_job_attr_delayed " + key + " " + value
+                    proc = subprocess.Popen(chirp_command.split(), shell = False, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+                    outs, errs = proc.communicate()
+                except Exception as e:
+                    proc.kill()
+                    print "An exception occurred: {}".format(e)
+                finally:
+                    if proc.returncode:
+                        print "WARNING::condor_chirp failed to set the attribute \'" + key + "\'"
```
kpedro88: I think this permissive approach is okay for now, but we might want to make it an exception at some point

aperloff:
This functionality can be taken advantage of by https://github.com/kpedro88/CondorProduction/pull/14.

TreeMaker/python/maker.py
aperloff: How do you want to keep track of this? Do you have a to-do list you want to put this on? Should I create an issue? Some other method of institutional memory?

kpedro88: i would probably just put a TODO comment in the code, but you can open an issue if you prefer

aperloff: A comment has been added to the code.

TreeMaker/python/maker.py
```diff
@@ -103,6 +106,31 @@ def __init__(self,parameters):
 
         self.readFiles = [(self.redir if val.startswith("/") else "")+val for val in self.readFiles]
 
+        # https://htcondor.readthedocs.io/en/latest/man-pages/condor_chirp.html?highlight=set_job_attr_delayed#chirp-commands
+        # https://github.com/cms-sw/cmssw/blob/master/FWCore/Services/plugins/CondorStatusUpdater.cc#L377-L410
+        try:
+            if os.environ["_CONDOR_CHIRP_CONFIG"]:
```
kpedro88: the outer try/catch seems unnecessary. why not just:
```python
chirp_env = "_CONDOR_CHIRP_CONFIG"
if chirp_env in os.environ and os.environ[chirp_env]:
    ...
```

aperloff: We check for the environment variable `_CONDOR_CHIRP_CONFIG` to see if we're on a condor worker node. However, when os.environ doesn't find an environment variable it returns a `KeyError`. We don't want that to stop the whole program, we just don't want to add a condor classad in that case. So the outer try catches that error, which we know will be coming from the one possible place which could generate a `KeyError`.

One solution I can come up with is to use:
```python
if os.environ.get("_CONDOR_CHIRP_CONFIG") is not None:
    <other stuff>
```
Would you like that better?

kpedro88: the point is that `os.environ` is just a `dict`. `if x in dict` doesn't throw an error. since `and` short-circuits, `os.environ[chirp_env]` will only be called if the key is present.

aperloff: Changed

PR #622, opened 2022-03-21
EMJ ultra legacy private production sample list

yimuchen:


yimuchen:
Delaying until EMJ MINIAOD files are moved to LPC space

aperloff:
Looks good to me.

PR #631, opened 2022-04-28
More speed improvements

kpedro88:
Tested on UL17 TTToSemiLeptonic, using the default workflow + tchannel (AK8 reclustering):
1. Clean out `GenLeptonRecoCand`: 0.4% faster
2. Remove pruned mass: 14% faster
3. Avoid copy: 0.3% faster
4. Puppi speedups (see https://github.com/cms-sw/cmssw/pull/37718): ~5% faster
5. Avoid duplicate genjet softdrop computation: 2% faster

Overall improvement in this workflow is 19.3%.

For comparison, the workflow with both tchannel & boosted (AK15 clustering) improves by 10.75%. The default workflow with no reclustering improves by 3.3%.

The majority of the improvement comes from disabling jet pruning in the AK8 clustering. The extent of this improvement suggests a similar possibility for SoftDrop (which is still used), which will be pursued next.


aperloff:
Looks good to me.
+1

PR #632, opened 2022-05-04
Combine base clustering and jet grooming algorithms

kpedro88:
I realized that `FastjetJetProducer` reruns the entire jet clustering, including the jet area computation, each time it is called (e.g. to get softdrop jets). This is computationally expensive and could be avoided by running the base clustering and multiple jet grooming algorithms all in the same instance of the producer. This PR adds to our setup a CMSSW branch (https://github.com/cms-sw/cmssw/compare/CMSSW_10_6_X...TreeMaker:OneShotTransforms106X) that modifies `FastjetJetProducer` (and `VirtualJetProducer`) accordingly, as well as a helper function to transform jet reclustering sequences to use the new, combined approach.

For the workflow with AK8 reclustering, this results in a 33% speedup. For the workflow with both AK8 and AK15 reclustering, this results in a 38% speedup. Similar speedups are expected for workflows using the "cleaned" reclustering for the Z->inv background estimation (which is no longer enabled by default).

Most of the speedup in the AK8 reclustering workflow is subsequently eliminated by reducing the `jetPtMin` cut for `SoftDrop` from the default 100 GeV to 5 GeV. This value should probably be tuned further, which will recover some percentage of the speedup. (The AK8+AK15 workflow still has a 31% speedup in this case.)

### DIFF
--- a/TreeMaker/python/makeTreeFromMiniAOD_cff.py
+++ b/TreeMaker/python/makeTreeFromMiniAOD_cff.py
@@ -385,7 +385,8 @@ def makeTreeFromMiniAOD(self,process):
             )
 
             # todo: revamp jet toolbox to handle this
-            process = self.transformJetSeq(process, "ak8GenJetsNoNu", {"SoftDrop":"ak8GenJetsNoNuSoftDrop"}, "recoGenJets")
+            if not self.doZinv: # calling jet toolbox again for ak8 would break after this, so skip it
+                process = self.transformJetSeq(process, "ak8GenJetsNoNu", {"SoftDrop":"ak8GenJetsNoNuSoftDrop"}, "recoGenJets")
             process.ak8PFJetsPuppiNoCutSoftDrop.jetPtMin = cms.double(5.0) # to match central clustering
             process = self.transformJetSeq(process, "ak8PFJetsPuppiNoCut", {"SoftDrop":"ak8PFJetsPuppiNoCutSoftDrop"}, "recoPFJets")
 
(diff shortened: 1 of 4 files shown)


kpedro88:
See https://github.com/cms-sw/cmssw/issues/37827 for more details

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -327,12 +365,13 @@ def makeTreeFromMiniAOD(self,process):
...
                 'ak8',
                 'jetSequence',
-                'out',
+                'noOutput',
                 PUMethod = 'Puppi',
                 useExistingWeights=True,
                 dataTier='miniAOD',
                 runOnMC = self.geninfo,
-                postFix = 'NoCut',
+                Cut = 'pt>10.',
```
aperloff, line 373: Why add the 10 GeV cut?

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -933,15 +978,16 @@ def makeTreeFromMiniAOD(self,process):
         from RecoJets.Configuration.RecoGenJets_cff import ak8GenJetsNoNu as ak8GenJetsNoNuDefault
         from RecoJets.JetProducers.SubJetParameters_cfi import SubJetParameters
         process.ak8GenJetsSoftDrop = ak8GenJetsNoNuDefault.clone(
-            useSoftDrop = cms.bool(True),
-            zcut = cms.double(0.1),
-            beta = cms.double(0.0),
-            R0   = cms.double(0.5),
-            useExplicitGhosts = cms.bool(True),
```
aperloff, line 940: You made this option non-configurable (default: `True`) and removed the configurations for the next option altogether. Was that on purpose?

TreeMaker/python/makeTreeFromMiniAOD_cff.py
kpedro88: See https://indico.cern.ch/event/1174262/contributions/4931776/attachments/2467601/4232118/treemaker%20UL%20performance%20jun%2021%202022.pdf

TreeMaker/python/makeTreeFromMiniAOD_cff.py
kpedro88: These options are now handled implicitly in my rewrite of `FastjetJetProducer`.

aperloff:
+1

PR #635, opened 2022-06-23
Faster constituent matching

kpedro88:
Using a hash table makes find O(1) rather than O(n) for linear search, at the cost of some overhead (creating the hash table and keeping it in memory). The current measured speedup is about 40% faster than the linear search version. The hash function may be able to be optimized further.

This PR also introduces a factory pattern (based on JetProperties) to request and store constituent properties. Currently, only PdgId is implemented. A few minor optimizations are propagated back to JetProperties.

### DIFF
--- a/TreeMaker/python/makeTreeFromMiniAOD_cff.py
+++ b/TreeMaker/python/makeTreeFromMiniAOD_cff.py
@@ -880,9 +880,9 @@ def makeTreeFromMiniAOD(self,process):
             'JetPropertiesAK8:jerFactorDown(JetsAK8_jerFactorDown)',
         ])
 
-    # record final jet collections
-    self.JetsTags.extend([JetTag,JetAK8Tag])
-    self.JetsNames.extend(["Jets","JetsAK8"])
+    # record final jet collections (excluding AK4)
+    self.JetsTags.extend([JetAK8Tag])
+    self.JetsNames.extend(["JetsAK8"])
 
     ## ----------------------------------------------------------------------------------------------
     ## GenJet variables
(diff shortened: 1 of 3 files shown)


kpedro88:
The updated data structure just uses the `edm::Ptr` `key()` property, assuming that the process and product indices are the same for all candidates in the jet (which is checked explicitly in the code, and is true for the collections in TreeMaker, though it may not be true for all conceivable cases). This is a substantially faster O(1) search, resulting in an 84% speedup over the original linear search version. See https://www.dropbox.com/s/9aqb6lrpgomjlal/treemaker%20UL%20performance%20jun%2026%202022.pdf?dl=0 for more details.

In making this update, I realized that the previous versions were not fully correct. They tried to match the AK8 and AK15 jet collections to `packedPFcandidates`, but these jets actually use the `puppipackedPFcandidates` collection (a clone of `packedPFcandidates` with the puppi weight applied). Since we usually only want to tag AK8 or larger jets, for now I have just removed AK4 from the list of collections used in this algorithm. (This also reduces the disk usage.) At some point, a more sophisticated approach could be adopted, as described in the code comments.

Since it was relevant for these checks, I added the puppi weight as a property that can be stored, though it is not currently stored by default.

aperloff:
+1

kpedro88:
Note for the future: https://github.com/cms-sw/cmssw/blob/master/FWCore/Utilities/interface/hash_combine.h may provide a better hash if it's necessary to return to the full hash table approach (e.g. because of constituents coming from multiple products).

PR #636, opened 2022-06-28
IP2D sign fix

yimuchen:
Added a new variable for the impact parameter sign. All information computed by default IPTools are kept as is. An additional `IPSign` variable is used to get the corrected sign of the IP variables.


aperloff:
+1

PR #642, opened 2022-07-21
[Run2_UL] Change copy protocol for T1_US_FNAL

aperloff:
Change the output protocol for T1_US_FNAL from gsiftp to WebDAV. gsiftp is being phased out and the default protocol for CMS transfers is now WebDAV. While this doesn't change the executable running the copy command (gfal-copy) it does change the protocol used.


aperloff:
This has been tested on CMS connect, making sure the jobs ran at T1_US_FNAL, and it works.

PR #644, opened 2022-09-13
Updates for pMSSM search [V17]

kpedro88:
* handle pMSSM scenario, model ID, file lists (some samples are not yet available and will be added later)
* add options to control output branches, in order to produce very slimmed-down ntuples for this specific search
  * bonus: for the most part, unscheduled mode will deactivate producers of discarded branches automatically, improving runtime
* added a separate option to disable the QG tagger for a minor speed improvement (this is an exception to the "bonus" above: unscheduled mode does not deactivate it because its output is accessed as userfloats of `pat::Jet`s)
* backported some minor technical speed improvements from V20

With the minimal set of output branches specified, the single-thread processing rate gets up to 90 evt/s (on an Intel(R) Xeon(R) W-2295 CPU @ 3.00GHz) and the output size is only 1.57 kb/evt.


TreeMaker/python/maker.py
```diff
@@ -56,12 +59,26 @@ def __init__(self,parameters):
         self.getParamDefault("applybaseline", False)
         self.getParamDefault("saveMinimalGenParticles", True)
 
+        # branch control options
+        self.getParamDefault("includeBranches", "")
+        self.getParamDefault("excludeBranches", "")
+        self.getParamDefault("exactBranches", False)
+        if len(self.includeBranches)>0 and len(self.excludeBranches)>0:
+            raise ValueError("includeBranches and excludeBranches are exclusive options; pick one!")
```
aperloff, line 67: Are these really exclusive if using regex? Can't you use a simpler regex to exclude most branches of a pattern, but include just a few? That is opposed to writing a really long regex with the inclusions excluded from what the regex matches.

TreeMaker/python/maker.py
```diff
@@ -56,12 +59,26 @@ def __init__(self,parameters):
...
 
+        # branch control options
+        self.getParamDefault("includeBranches", "")
+        self.getParamDefault("excludeBranches", "")
+        self.getParamDefault("exactBranches", False)
+        if len(self.includeBranches)>0 and len(self.excludeBranches)>0:
+            raise ValueError("includeBranches and excludeBranches are exclusive options; pick one!")
+        for branchListName in ["includeBranches","excludeBranches"]:
+            branches = getattr(self,branchListName)
+            if not isinstance(branches,list) and branches.endswith(".txt"):
```
aperloff, line 70: Why not make the list a python list in a file which can be imported? Parsing a text file just seems like an added complication, not that it's super hard (obviously).

TreeMaker/python/maker.py
```diff
@@ -138,9 +149,11 @@ def printSetup(self):
         print " Instance name of tag information: "+self.tagname
         print " Instance name of HLT tag information: "+self.hlttagname
         print " Including gen-level information: "+str(self.geninfo)
-        print " Including pMSSM-related information: "+str(self.pmssm)
         print " Using fastsim settings: "+str(self.fastsim)
         print " Running signal uncertainties: "+str(self.signal)
+        if len(self.includeBranches)>0: print " Including branches: "+','.join(self.includeBranches)
+        if len(self.excludeBranches)>0: print " Excluding branches: "+','.join(self.excludeBranches)
+        if self.exactBranches: print "  Using exact branches (no regex)"
```
aperloff, line 156: I feel like there should be an easier way to automatically detect if something is a regex, but I can't think of it right now. In general, it would be nice to not have a manual switch and to be able to mix exact patterns with regex.

Utils/python/pfnInPath.py
```diff
@@ -0,0 +1,18 @@
+# from https://github.com/cms-sw/cmssw/blob/master/FWCore/ParameterSet/python/pfnInPath.py
```
aperloff, line 1: Why does this need to be included in TreeMaker? Why can't we import the CMSSW version?

Production/test/condorSub/dict_Autumn18Fast_pmssm.py
```diff
@@ -0,0 +1,11 @@
+flist = {
+    "scenario": "Autumn18Fastsig",
```
aperloff, line 2: Wouldn't you want an args section in these dictionaries to turn on things like QGLikelihood or are there no arguments needed?

Production/test/condorSub/dict_Fall17Fast_pmssm.py
```diff
@@ -0,0 +1,12 @@
+flist = {
+    "scenario": "Fall17Fastsig",
```
aperloff, line 2: Same args comment here.

TreeMaker/python/maker.py
kpedro88: But the order becomes ambiguous: the user would need to specify whether include or exclude comes first. Then one could imagine even more complicated use cases, but I don't see a motivation to support them right now.

TreeMaker/python/maker.py
kpedro88: In this case, I don't think importing an arbitrary python file is easier than importing a text file (which is intended to have exactly one branch name per line).

TreeMaker/python/maker.py
kpedro88: Again, not sure I see the motivation for such a complicated use case at this time.

Utils/python/pfnInPath.py
kpedro88: The CMSSW version is not available in 10_2_X.

Production/test/condorSub/dict_Autumn18Fast_pmssm.py
kpedro88: The V17 production does not use args in dictionaries.

Production/test/condorSub/dict_Fall17Fast_pmssm.py
kpedro88: See above

Utils/python/pfnInPath.py
aperloff: Ah, okay. I must have only checked 10_6_X. This file should be removed whenever these changes get forward ported to the UL branch.

aperloff:
+1

PR #652, opened 2022-09-30
Adding pileupIDs for AK4 jets

kapsiak:
Adding pileup ids for AK4 jets

### DIFF
--- a/TreeMaker/python/makeJetVars.py
+++ b/TreeMaker/python/makeJetVars.py
@@ -59,7 +59,7 @@ def makeGoodJets(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInpu
 # 0 = scalars (+ origIndex,jerFactor for syst)
 # 1 = 0 + 4vecs, masks, minimal set of properties
 # 2 = all properties
-def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInputTag(), onlyGoodJets=False, systType=""):
+def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInputTag(), onlyGoodJets=False, systType="", storePileupIds=False):
     ## ----------------------------------------------------------------------------------------------
     ## GoodJets
     ## ----------------------------------------------------------------------------------------------
@@ -149,6 +149,8 @@ def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInput
     ## ----------------------------------------------------------------------------------------------
     ## Jet properties
     ## ----------------------------------------------------------------------------------------------
+
+
     if storeProperties==0:
         # for systematics
         JetProperties = cms.EDProducer("JetProperties",
@@ -171,6 +173,28 @@ def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInput
         JetProperties = jetproperties.clone(
             JetTag       = GoodJetsTag
         )
+
+        if storePileupIds:
+            from TreeMaker.Utils.pileupjetid_cfi import pileupJetId, pileupjetalgos
+            pileupJetIdUpdated = pileupJetId.clone(
+                jets = GoodJetsTag,
+                inputIsCorrected=True,
+                applyJec=False,
+                vertexes = cms.InputTag("offlineSlimmedPrimaryVertices"),
+                algos=pileupjetalgos["2016UL"]
+            )
+            (TMeras.TMUL2016APV).toModify(pileupJetIdUpdated,algos = pileupjetalgos["2016UL_APV"])
+            (TMeras.TMUL2017).toModify(pileupJetIdUpdated,algos = pileupjetalgos["2017UL"])
+            (TMeras.TMUL2018).toModify(pileupJetIdUpdated,algos = pileupjetalgos["2018UL"])
+
+            setattr(process,"pileupJetId"+suff,pileupJetIdUpdated)
+
+            JetProperties.properties.extend(["pileupJetId"])
+            JetProperties.pileupJetId = cms.vstring('pileupJetId' + suff + ':fullDiscriminant')
+            self.VectorDouble.extend([
+                'JetProperties'+suff+':pileupJetId(Jets'+suff+'_pileupId)'
+            ])
+
 
         # provide extra info where necessary
         if storeProperties==1:
--- a/TreeMaker/python/makeJetVars.py
+++ b/TreeMaker/python/makeJetVars.py
@@ -59,7 +59,7 @@ def makeGoodJets(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInpu
 # 0 = scalars (+ origIndex,jerFactor for syst)
 # 1 = 0 + 4vecs, masks, minimal set of properties
 # 2 = all properties
-def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInputTag(), onlyGoodJets=False, systType="", storePileupIds=False):
+def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInputTag(), onlyGoodJets=False, systType=""):
     ## ----------------------------------------------------------------------------------------------
     ## GoodJets
     ## ----------------------------------------------------------------------------------------------
@@ -149,8 +149,6 @@ def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInput
     ## ----------------------------------------------------------------------------------------------
     ## Jet properties
     ## ----------------------------------------------------------------------------------------------
-
-
     if storeProperties==0:
         # for systematics
         JetProperties = cms.EDProducer("JetProperties",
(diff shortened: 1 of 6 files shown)


TreeMaker/python/makeJetVars.py
```diff
@@ -149,6 +149,8 @@ def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInput
     ## ----------------------------------------------------------------------------------------------
     ## Jet properties
     ## ----------------------------------------------------------------------------------------------
+
+
```
kpedro88: remove unnecessary newlines

Utils/src/JetProperties.cc
```diff
@@ -247,6 +262,8 @@ DEFAULT_NAMED_PTR(bDiscriminator,pfMassIndependentDeepDoubleBvLJetTagsProbHbb);
 DEFAULT_NAMED_PTR(bDiscriminator,pfMassIndependentDeepDoubleCvLJetTagsProbHcc);
 DEFAULT_NAMED_PTR(bDiscriminator,pfMassIndependentDeepDoubleCvBJetTagsProbHcc);
 
+
+
```
kpedro88: remove unnecessary newlines

TreeMaker/python/makeJetVars.py
```diff
@@ -172,6 +174,28 @@ def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInput
             JetTag       = GoodJetsTag
         )
 
+        if storePileupIds:
+            from TreeMaker.Utils.pileupjetid_cfi import pileupJetId, pileupjetalgos
```
kpedro88, line 171: I think you neglected to do `git add TreeMaker/Utils/python/pileupjetid_cfi.py`

Utils/src/JetProperties.cc
```diff
@@ -127,6 +127,21 @@ DEFAULT_NAMED_PTR(D,neutralHadronPuppiMultiplicity);
...
+	public:
+		using NamedPtr<double>::NamedPtr;
+		//default for user floats
+		void get_property(const pat::Jet& Jet) override { 
+                    if(Jet.pt() < 50.0 ){
+                        push_back(Jet.userFloat(extraInfo.at(0)));
+                    } else {
+                        push_back(-HUGE_VAL);
+                    }
+                }
```
kpedro88: fix the indentation here to match the rest of the file

TreeMaker/python/makeJetVars.py
```diff
@@ -172,6 +174,28 @@ def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInput
...
+                algos=pileupjetalgos["2016UL"]
+            )
+            (TMeras.TMUL2016APV).toModify(pileupJetIdUpdated,algos = pileupjetalgos["2016UL_APV"])
+            (TMeras.TMUL2017).toModify(pileupJetIdUpdated,algos = pileupjetalgos["2017UL"])
+            (TMeras.TMUL2018).toModify(pileupJetIdUpdated,algos = pileupjetalgos["2018UL"])
+
+            setattr(process,"pileupJetId"+suff,pileupJetIdUpdated)
+
+            JetProperties.properties.extend(["pileupJetId"])
+            JetProperties.pileupJetId = cms.vstring('pileupJetId' + suff + ':fullDiscriminant')
```
kpedro88: have you tested that this actually works? the jet collection is never updated to include this userfloat...

kpedro88:
in addition to the review comments, I propose adding a flag to disable this in case someone is running a test or mini-ntuple and doesn't need it. you can follow https://github.com/TreeMaker/TreeMaker/commit/030691dcf11b7f4640ec674a9455ffc11fadd0c1 to see how I recently did this for a similar case

kapsiak:
Ok, while I work on these, one question is the method of pileup subtraction currently used in the jets. The PileUpID twiki page https://twiki.cern.ch/twiki/bin/view/CMS/PileupJetIDUL seems to suggest these corrections should be applied to CHS jets, are these what are used in the current NTuples?

Secondly, I was hoping if you could check (I am not so experienced with CMSSW) that the pileup IDs will be run on JEC corrected jets.

Utils/src/JetProperties.cc
```diff
@@ -127,6 +127,21 @@ DEFAULT_NAMED_PTR(D,neutralHadronPuppiMultiplicity);
 DEFAULT_NAMED_PTR(D,photonPuppiMultiplicity);
 DEFAULT_NAMED_PTR(D,msd); // softdrop mass as userfloat rather than from subjets
 
+class NamedPtr_pileupId : public NamedPtr<double> {
```
kpedro88: move this down to where the other specialized helper classes are defined below

Utils/src/JetProperties.cc
```diff
@@ -127,6 +127,21 @@ DEFAULT_NAMED_PTR(D,neutralHadronPuppiMultiplicity);
...
+		void get_property(const pat::Jet& Jet) override { 
+                        if(Jet.pt() < 50.0 ){
+                                push_back(Jet.userFloat(extraInfo.at(0)));
+                        } else {
+                                push_back(-HUGE_VAL);
+                        }
+                }
+};
+
+DEFAULT_NAMED_PTR(pileupId,pileupJetId);
```
kpedro88: use `DEFINE_NAMED_PTR()` as is done for the other specialized helper classes below

kpedro88:
Yes, the default AK4 jets for Run 2 are CHS jets.

kpedro88:
Yes, the JECs are applied to the jet collection before you call the PileupJetID module.

kpedro88:
needs to be rebased

Utils/src/JetProperties.cc
```diff
@@ -211,6 +212,20 @@ class NamedPtr_hfHadronEnergyFraction : public NamedPtr<double> {
...
+		//default for user floats
+		void get_property(const pat::Jet& Jet) override { 
+                        if(Jet.pt() < 50.0 ){
+                                push_back(Jet.userFloat(extraInfo.at(0)));
+                        } else {
+                                push_back(-HUGE_VAL);
+                        }
+                }
+};
+DEFINE_NAMED_PTR(pileupId,pileupJetId);
```
kpedro88: DEFINE_NAMED_PTR only takes one argument

kpedro88:
and some review comments have not been addressed

kapsiak:
Sorry, would you like me to rebase the Run2UL branch to the PileupId branch?

kpedro88:
You need to rebase your branch on top of the Run2_UL branch.

TreeMaker/python/makeJetVars.py
kapsiak: I have run some of the tests and it does produce a filled branch in the NTuple. One point I wanted to ask is about the use of HUGE_VAL for producing an "invalid value," since PileupID is only valid for jets with pt<50, and produces a meaningless value for pt>50. Perhaps better would be std::optional<float> (if TreeMaker supports it). Or I could also just provide no check the user would need to enforce this in their analysis framework. Please let me know what you think.

TreeMaker/python/makeJetVars.py
kpedro88: still not removed

TreeMaker/python/makeJetVars.py
kpedro88: I suspect it is just picking up the default userfloat from the miniAOD, since your new userfloat is not included in any jet collections. If what is in the miniAOD uses the correct training, then `pileupJetIdUpdated` is not needed here and can just be removed. If what is in the miniAOD does not use the correct training, then you need to update the jet collection with the new userfloat.

TreeMaker/python/makeJetVars.py
```diff
@@ -59,7 +59,7 @@ def makeGoodJets(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInpu
 # 0 = scalars (+ origIndex,jerFactor for syst)
 # 1 = 0 + 4vecs, masks, minimal set of properties
 # 2 = all properties
-def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInputTag(), onlyGoodJets=False, systType=""):
+def makeJetVars(self, process, JetTag, suff, storeProperties, SkipTag=cms.VInputTag(), onlyGoodJets=False, systType="", storePileupIds=False):
```
kpedro88: change to this line is no longer needed since a member variable is now being used

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -859,6 +859,7 @@ def makeTreeFromMiniAOD(self,process):
         suff='',
         storeProperties=2,
         SkipTag=SkipTag,
+        storePileupIds=True
```
kpedro88: change to this line is no longer needed since a member variable is now being used

TreeMaker/python/makeJetVars.py
kapsiak: Hi Kevin. Sorry, I am not very familiar with this, but the twiki seems to suggest that the discriminant can be accessed through the userfloat. https://twiki.cern.ch/twiki/bin/view/CMS/PileupJetIDUL . I am not sure how to add it manually.

kapsiak: Ok, I believe I have resolved this issue. As a check I have the histograms for unittest 2 before and after commit [ad0d760](https://github.com/TreeMaker/TreeMaker/pull/652/commits/ad0d7607749273fff80f931e6fa163d37bd145f5).

Pileup_Pre_PR
ved this issue. 
PileUpID_Post_PR

PR #653, opened 2022-09-30
Changing Muon effective area values to 94X

kapsiak:


kpedro88:
This looks fine, but because it changes ntuple contents (rather than just adding new branches), we will leave it unmerged until the plan for the ntuple update is finalized.

PR #670, opened 2022-12-21
Handle PDFs, Scales, and PSs for Stealth Signals

JHiltbrand:
Ultra Legacy Stealth signal samples do not contain any PDF sets (`LHEEventProduct` is missing from the event). Likewise, the information contained in the `GenEventInfoProduct` pertains to factorization/renormalization weights, _and_ PS weights.

Thus, this PR allows `TreeMaker` to accommodate such a scenario for the Stealth signals, while still trying to be compatible with all other scenarios. The chosen PDF set is based on the recommendations at [1], for positive-definite sets.

[1] https://cms-pdmv.gitbook.io/project/mccontact/info-for-mc-production-for-ultra-legacy-campaigns-2016-2017-2018#recommendations-on-the-usage-of-pdfs-and-cpx-tunes

### DIFF
--- a/Utils/src/PDFWeightProducer.cc
+++ b/Utils/src/PDFWeightProducer.cc
@@ -158,6 +158,7 @@ void PDFWeightProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::Ev
   bool found_scales = false;
   bool found_pdfs = false;
   bool found_pss = false;
+  unsigned offset = 0;
   
   if(!handles.empty()){
     edm::Handle<LHEEventProduct> LheInfo = handles[0];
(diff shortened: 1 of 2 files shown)


Utils/src/PDFWeightProducer.cc
```diff
@@ -181,9 +181,16 @@ void PDFWeightProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::Ev
...
+    else if (!found_pss and nPSs_>0){
+        unsigned offset = 0;
+        found_pss = helper.fillWeights(psweights.get(),offset,nPSs_,wtype::ps);
+    }
+
+    // Extra check here when scales and PS weights are in GenEventInfoProduct
+    // Scales should have just been found above, but still need to go for PS now
+    if(!found_pss and found_scales and nPSs_>0){
+      unsigned offset = 1;
+      found_pss = helper.fillWeights(psweights.get(),nScales_+offset,nPSs_,wtype::ps);
```
kpedro88: to avoid having to check for PS weights twice (while still handling the more common case where `LHEEventProduct` contains scales & pdfs while `GenEventInfoProduct` just contains PS weights), i would move the definition of `unsigned offset = 0` up to L160 (where the `found_` variables are declared), and then on L182, the offset can be set to `1+nScales` (i.e. only use that offset if the scale weights are found specifically in the `GenEventInfoProduct`).

JHiltbrand: How about [f3ab8a1](https://github.com/TreeMaker/TreeMaker/pull/670/commits/f3ab8a13e90bba68ce44e48376964520b627a44d) ?

Utils/src/PDFWeightProducer.cc
```diff
@@ -175,14 +176,18 @@ void PDFWeightProducer::produce(edm::StreamID, edm::Event& iEvent, const edm::Ev
   if(genHandle.isValid()){
     auto helper = makeHelper(genHandle.product(),norm_);
     if((!found_scales and nScales_>0) or (!found_pdfs and nPDFs_>0)){
-      unsigned offset = 1;
       //renormalization/factorization scale weights
+      offset++;
```
kpedro88, line 180: doesn't work for the more common case described in the previous comment. `offset` should be 0 when going into the `if(!found_pss...)` block, *unless* `found_scales` is true. in the current commit, `offset` will be (at least) 1 going into that block all the time. hence the suggestion to check the status of `found_scales` *inside* the `if(!found_pss...)` block when determining the offset.

JHiltbrand: Hmm, I figured for the common case, we get both the PDFs and scales from the `LHEEventProduct` and subsequently, we would _never_ enter the block at L178 - thereby leaving `offset` at its initialized value of 0 upon going into the `if(!found_pss...)` block to get the PSs.
Conversely, for the stealth signal case, we'll go in at L178 and `offset` will come out as `nScales_+1` for use in the PS block. Sorry if I am missing something obvious here.

kpedro88: Ah sorry, the case where it doesn't work is Pythia-only signals, where the only weights provided are PS weights in the `GenEventInfoProduct`. Those samples will still go into this block and increment the offset incorrectly.

JHiltbrand: Gotcha, alright, how about [6eb19af](https://github.com/TreeMaker/TreeMaker/pull/670/commits/6eb19af2c1c89e55c1e9f085eb282a28427343d9) ?

PR #672, opened 2023-01-30
Gen weights by name + code improvements

kpedro88:
Apparently the UL generator weights have no consistent order, so we have to get them by name. The order of names specified in the `PDFWeightProducer` config is the order that will be output in TreeMaker, which preserves consistency with pre-L.

This PR also incorporates some code improvements:
1. Also use a LumiCache for SignalScanProducer (marginally reduces memory usage vs. stream producer)
2. Split weight recalculations, which are not thread-safe, into a separate one producer instead of using mutexes (should be marginally more efficient to schedule tasks in multithreading mode, according to framework developers).

Tested on `Summer20UL18.StealthSYY_2t6j_mStop-300to1400_mSo-100_TuneCP5_13TeV-madgraphMLM-pythia8`, but more tests welcome...


kpedro88:
Profiling with 4 threads and 10K events from the above sample shows that the PDF weight modules are about 15% faster now.

JHiltbrand:
I tested with `Summer20UL16.TTToSemiLeptonic_TuneCP5_13TeV-powheg-pythia8` and did a "before-and-after". The PDF and scale weights stored in the output ntuple do not change when using your code changes - as is expected. Likewise, I see that the 14 stored PSweights do change reasonably i.e. picking up different values from the array of 45 elements.

PR #676, opened 2023-05-22
Gen weight refactor

kpedro88:
There are too many variations of generator weights (scale, PDF, parton shower) to continue to handle manually on a case-by-case basis. Instead, centrally developed parsing code is now used.

(The "normalize" option, rarely used anyway, is removed for simplicity. Users can normalize to the first weight in the vector, usually 1 anyway, offline if desired.)

This PR also fixes the handling of weights for MadGraph signals (as opposed to existing settings, which were for Pythia-only signals).

Attn: @tklijnsma @JHiltbrand


JHiltbrand:
Hello @kpedro88,

I ran a couple of very basic checks, where I ran unit tests with the changes introduced in the last GEN weights fix, and then I ran with your changes in this PR here. The two samples I ran on were 

- `Summer20UL16.TTToSemiLeptonic_TuneCP5_13TeV-powheg-pythia8`
- `Summer20UL17.RPV_2t6j_mStop-300to1400_mN1-100_TuneCP5_13TeV-madgraphMLM-pythia8`.

In comparing the `PDFweights`, `PSweights`, and `ScaleWeights` branches between the pairs of ROOT files, I observe identical entries.

PR #695, opened 2024-03-22
Adding config, dict for ZLLGamma samples 

alpana-hep:
This pull request is adding the configuration files for ZLL gamma sample of 2018,2017,2016,2016APV. Also the dictionaries to submit condor jobs for the same. The cross section for the process has also been added along with effective events to get effective weights for events with negative weights.


kpedro88:
Can you remove `Production/python/Summer20UL16APV/ZLLGJets_MonoPhoton_PtG-15to130_TuneCP5_13TeV-amcatnloFXFX-pythia8_cff.py` and the reference to it in `Production/test/condorSub/dict_Summer20UL16APV_ZtoLLgjets.py`? Since it is a) the wrong miniAOD version and b) not available on disk, so I don't think we can/should use it.

alpana-hep:
Yes, totally agree. I missed it. Removing these now

PR #706, opened 2025-05-08
Clean branch with Lund Reweighting Code 

easmith5:
This branch matches gen-reco particles and cluster them into constituents of specific dark hadrons. These can then be reweighting using the Lund Reweighting code.


TreeMaker/interface/TreeMaker.h
```diff
@@ -219,9 +219,10 @@ typedef TreeObject<vector<math::PtEtaPhiELorentzVector>,vector<math::PtEtaPhiELo
 typedef TreeObject<vector<math::XYZVector>,vector<math::XYZVectorF>> TreeObjectVXYZVToF;
 typedef TreeObject<vector<math::XYZPoint>,vector<math::XYZPointF>> TreeObjectVXYZPToF;
 typedef TreeObject<vector<vector<math::PtEtaPhiELorentzVector>>,vector<vector<math::PtEtaPhiELorentzVectorF>>> TreeObjectVVLVToF;
+typedef TreeObject<vector<vector<vector<math::PtEtaPhiELorentzVector>>>,vector<vector<vector<math::PtEtaPhiELorentzVectorF>>>> TreeObjectVVVLVToF;
 typedef TreeObject<vector<vector<math::XYZVector>>,vector<vector<math::XYZVectorF>>> TreeObjectVVXYZVToF;
 typedef TreeObject<vector<vector<math::XYZPoint>>,vector<vector<math::XYZPointF>>> TreeObjectVVXYZPToF;
-//typedefs for doubles
+//typedefs for doublesx
```
kpedro88: revert typo?

TreeMaker/interface/TreeMaker.h
```diff
@@ -495,7 +512,7 @@ class TreeNestedVector : public TreeObject<std::vector<std::vector<BaseIn>>,std:
 		}
 		void SetConsumes(edm::ConsumesCollector && iC) override{
 			tok = iC.consumes<TopIn>(this->tag);
-		}
+ 		}
```
kpedro88: revert random space?

TreeMaker/interface/TreeMaker.h
```diff
@@ -594,3 +611,123 @@ template <>
...
+			if (!associated && storeOffsets) { offsets.insert(std::end(offsets),0); }
+			for(auto& sub : all) {
+			    int subLength = 0;
+			    for(auto& isub : sub) {
+			        subLength += isub.size();
+			    }
+			    if (!associated && storeOffsets) { offsets.insert(std::end(offsets),subLength); }
+			    for(auto& isub : sub) {
+			        accum.insert(std::end(accum), std::begin(isub), std::end(isub));
+				if (!associated) offsets.insert(std::end(offsets), storeOffsets ? accum.size() : isub.size());
```
kpedro88: fix inconsistent usage of tabs vs. spaces for indentation (everywhere)

TreeMaker/interface/TreeMaker.h
```diff
@@ -594,3 +611,123 @@ template <>
...
+		TreeNNVector(string tempFull_, string title_="", bool nestedVectors_=true, bool storeOffsets_=true, bool associated_=false, int splitLevel_=0) :
+			TreeObject<TripIn,TripOut>(tempFull_,title_,splitLevel_), nestedVectors(nestedVectors_), storeOffsets(storeOffsets_), associated(associated_) {}
+		// Destructor
+		~TreeNNVector() override {}
+		
+		// Functions
+		// From: https://stackoverflow.com/questions/17294629/merging-flattening-sub-vectors-into-a-single-vector-c-converting-2d-to-1d
+		void flatten(TripIn const& all, SubOut &accum, vector<int> &offsets) {
+			// Don't store any offsets if there are no sub-vectors
+			if (all.size() == 0) return;
```
kpedro88: `if (all.empty())` (more efficient)

TreeMaker/interface/TreeMaker.h
```diff
@@ -594,3 +611,123 @@ template <>
...
+			    }
+			    if (!associated && storeOffsets) { offsets.insert(std::end(offsets),subLength); }
+			    for(auto& isub : sub) {
+			        accum.insert(std::end(accum), std::begin(isub), std::end(isub));
+				if (!associated) offsets.insert(std::end(offsets), storeOffsets ? accum.size() : isub.size());
+			    }
+			}
+			// This protects against the case where there were >=1 empty sub-vectors, and only empty vectors
+			// Thus, nothing will be in the output (accum) vector, but the offsets would be all '0'
+			if (storeOffsets && !associated && accum.size() == 0) offsets.clear();
```
kpedro88: `accum.empty()`

TreeMaker/interface/TreeMaker.h
```diff
@@ -594,3 +611,123 @@ template <>
...
+			    for(auto& isub : sub) {
+			        subLength += isub.size();
+			    }
+			    if (!associated && storeOffsets) { offsets.insert(std::end(offsets),subLength); }
+			    for(auto& isub : sub) {
+			        accum.insert(std::end(accum), std::begin(isub), std::end(isub));
+				if (!associated) offsets.insert(std::end(offsets), storeOffsets ? accum.size() : isub.size());
+			    }
+			}
+			// This protects against the case where there were >=1 empty sub-vectors, and only empty vectors
```
kpedro88, line 656: to clarify this comment a little: it's the case where *everything* is empty *and* there are >=1 sub-vectors, right?

TreeMaker/interface/TreeMaker.h
kpedro88: (and please follow existing usage in each file, i.e. don't change indentation in other lines not already modified)

TreeMaker/interface/TreeMaker.h
```diff
@@ -594,3 +611,123 @@ template <>
...
+			}
+			else{
+				values.clear();
+				if (!associated) offsets.clear();
+			}
+		}
+		const string GetTripType() {
+			return "vector<" + GetTopType() + ">";
+		}
+                const string GetTopType() {
```
kpedro88: more spacing/indentation issues

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -1180,7 +1184,19 @@ def makeTreeFromMiniAOD(self,process):
             DarkHadronIDs = cms.vuint32(4900111,4900113,4900211,4900213),
             DarkGluonIDs = cms.vuint32(4900021),
             SMQuarkIDs = cms.vuint32(1,2,3,4,5,6),
+            nsubjettiness = cms.PSet(
+                useSoftDrop = cms.bool(True),
```
kpedro88: this parameter is not actually used for nsubjettiness, can be removed

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -1180,7 +1184,19 @@ def makeTreeFromMiniAOD(self,process):
...
+                Rcutoff = cms.double(999.0),
+                beta = cms.double(1.0),
+                R0   = cms.double(0.8),
+                measureDefinition = cms.uint32(0),
+                axesDefinition = cms.uint32(6),
+                nPass = cms.int32(999),
+                akAxesR0 = cms.double(999.0),
+            ),
+            #jetCollInstanceName=cms.string("SubJets"),
```
kpedro88: delete commented-out code

TreeMaker/python/makeTreeFromMiniAOD_cff.py
```diff
@@ -1197,9 +1213,41 @@ def makeTreeFromMiniAOD(self,process):
...
+                'HiddenSector:GenJetsDarkHadrons(GenJetsAK8_darkHadrons)',
+                'HiddenSector:GenJetsDarkHadronJets(GenJetsAK8_darkHadronJets)',
+                'HiddenSector:JetsDarkHadronJets(JetsAK8_darkHadronJets)',
+            ])
+            self.VectorVectorVectorLorentzVector.extend([
+                'HiddenSector:GenJetsDarkHadronJetsConstituents(GenJetsAK8_darkHadronJets_constituents)',
+                'HiddenSector:JetsDarkHadronJetsConstituents(JetsAK8_darkHadronJets_constituents)',
+                'HiddenSector:JetsDarkHadronJetsConstituentsGenMatchOnly(JetsAK8_darkHadronJets_constituentsGenMatchOnly)',
+                'HiddenSector:JetsDarkHadronJetsConstituentsNextDH(JetsAK8_darkHadronJets_constituentsNextDH)',
+
```
kpedro88: delete extra line (also below)

TreeMaker/src/TreeMaker.cc
```diff
@@ -129,6 +129,7 @@ TreeMaker::TreeMaker(const edm::ParameterSet& iConfig) :
 				case TreeTypes::t_vvdouble  : tmp = saveFloat ? (TreeObjectBase*)(new TreeNVDoubleToF(VarName,VarTitle,nestedVectors,storeOffsets,false,splitLevel)) : (TreeObjectBase*)(new TreeNVDouble(VarName,VarTitle,nestedVectors,storeOffsets,false,splitLevel)); break;
 				case TreeTypes::t_vvstring  : tmp = new TreeNVString(VarName,VarTitle,nestedVectors,storeOffsets,false,splitLevel); break;
 				case TreeTypes::t_vvlorentz : tmp = saveFloat ? (TreeObjectBase*)(new TreeNVLVToF(VarName,VarTitle,nestedVectors,storeOffsets,false,splitLevel)) : (TreeObjectBase*)(new TreeNVLV(VarName,VarTitle,nestedVectors,storeOffsets,false,splitLevel)); break;
+			        case TreeTypes::t_vvvlorentz: tmp = saveFloat ? (TreeObjectBase*)(new TreeNNVLVToF(VarName,VarTitle,nestedVectors,storeOffsets,false,splitLevel)) : (TreeObjectBase*)(new TreeNNVLV(VarName,VarTitle,nestedVectors,storeOffsets,false,splitLevel)); break;
```
kpedro88: tabs vs spaces

Utils/src/HiddenSectorProducer.cc
```diff
@@ -40,37 +44,50 @@ CandPtr daughter_noexcept(const T& mother, unsigned i) {
...
     edm::EDGetTokenT<edm::View<reco::GenMET>> GenMetTok_;
     edm::EDGetTokenT<edm::View<reco::GenParticle>> GenTok_;
     edm::EDGetTokenT<edm::View<reco::GenJet>> GenJetTok_;
+    edm::EDGetTokenT<edm::View<reco::Candidate>> CandTok_;
+    edm::EDGetTokenT<std::vector<int>> GenIndexTok_;
     double coneSize_;
+    edm::ParameterSet njhConfig; 
     PidSet DarkSMediatorIDs_, DarkTMediatorIDs_, DarkQuarkIDs_, DarkHadronIDs_, DarkGluonIDs_, DarkStableIDs_, DarkFirstIDs_, SMQuarkIDs_;
+    std::unique_ptr<NjettinessHelper> beginStream(edm::StreamID) const {
+      return std::unique_ptr<NjettinessHelper>(new NjettinessHelper(njhConfig));
```
kpedro88: prefer `return std::make_unique<NjettinessHelper>(njhConfig);` (more concise, avoids `new ` which always looks like a memory leak even though it's not in this case)

Utils/src/HiddenSectorProducer.cc
```diff
@@ -40,37 +44,50 @@ CandPtr daughter_noexcept(const T& mother, unsigned i) {
...
+    }
+
 };
 
-void HiddenSectorProducer::fillSet(PidSet& IDset, const std::string& name, const edm::ParameterSet& iConfig){
+void HiddenSectorProducer::fillSet(PidSet& IDset, const std::string& name, const edm::ParameterSet& iConfig)
+{
```
kpedro88, line 92: unnecessary change

Utils/src/HiddenSectorProducer.cc
```diff
@@ -196,19 +237,166 @@ double HiddenSectorProducer::calculateMT2(const edm::Handle<edm::View<reco::GenM
...
+
+    for(unsigned j = 0; j < h_jets->size(); ++j){
+      if (cands_out.size() <= j) cands_out.emplace_back();
+      if (pdgids_out.size() <= j) pdgids_out.emplace_back();
+      const auto& jet = h_jets->at(j);
+      const auto& daughterPtrs = jet.daughterPtrVector();
+      if(cands_out[j].size()==daughterPtrs.size()) continue;
+      bool keep = false; 
+      for(const auto& candJetDau : jet.daughterPtrVector()){ 
+	keep = candJetDau.key() == candPtr.key();
```
kpedro88: tabs vs spaces

Utils/src/HiddenSectorProducer.cc
```diff
@@ -196,19 +237,166 @@ double HiddenSectorProducer::calculateMT2(const edm::Handle<edm::View<reco::GenM
...
+
+  //loop over PF candidate collection once: check every jet in every jet collection
+  //only PF candidates found in a jet collection will be kept
+  for(unsigned c = 0; c < h_cands->size(); ++c){
+    const auto& candPtr = h_cands->ptrs()[c];
+    //if this cand is kept, it will be appended to cands_out
+
+    for(unsigned j = 0; j < h_jets->size(); ++j){
+      if (cands_out.size() <= j) cands_out.emplace_back();
+      if (pdgids_out.size() <= j) pdgids_out.emplace_back();
```
kpedro88: could one just do `resize(j)` for these two `out` vectors at the beginning of the function (outside of the loop) rather than emplacing one at a time?

Utils/src/HiddenSectorProducer.cc
```diff
@@ -196,19 +237,166 @@ double HiddenSectorProducer::calculateMT2(const edm::Handle<edm::View<reco::GenM
...
+	int index = it - pdgids.begin(); 
+	toMatch[index].genParts.push_back(genSubjetConstituents[i][j]);
+	toMatch[index].genIndexTop.push_back(i);
+	toMatch[index].genIndex.push_back(j);
+      }
+      else if (std::find(pdgids.begin(), pdgids.end(), genSubjetPdgid[i][j]) == pdgids.end() ) { 
+	toMatch[pdgids.size()].genParts.push_back(genSubjetConstituents[i][j]);
+	toMatch[pdgids.size()].genIndexTop.push_back(i);
+	toMatch[pdgids.size()].genIndex.push_back(j);
+      }
```
kpedro88: further reduce duplication:
```cpp
int index = -1;
if (it != pdgids.end()) index = it - pdgids.begin();
else index = pdgids.size();
toMatch[index]...
```

Utils/src/HiddenSectorProducer.cc
```diff
@@ -196,19 +237,166 @@ double HiddenSectorProducer::calculateMT2(const edm::Handle<edm::View<reco::GenM
...
+    auto it = std::find(pdgids.begin(), pdgids.end(), jetCandsPdgid[r]);
+    if(it != pdgids.end() ) {
+      int index = it - pdgids.begin();
+      toMatch[index].recoParts.push_back(jetCands[r]);
+      toMatch[index].recoIndex.push_back(r);
+    }
+    else if(std::find(pdgids.begin(), pdgids.end(), jetCandsPdgid[r]) == pdgids.end() ) {
+      toMatch[pdgids.size()].recoParts.push_back(jetCands[r]);
+      toMatch[pdgids.size()].recoIndex.push_back(r);
+    }
```
kpedro88: similarly here

Utils/src/HiddenSectorProducer.cc
```diff
@@ -196,19 +237,166 @@ double HiddenSectorProducer::calculateMT2(const edm::Handle<edm::View<reco::GenM
...
+
+  }
+
+  for( auto& m: toMatch ) {  
+    m.matchedIndex = utils::matchAB(m.genParts, m.recoParts);
+    //Save matched indices to final array
+    for(unsigned i = 0; i < m.matchedIndex.size(); i++){
+      if (m.matchedIndex[i] != -1) {
+	indices[m.genIndexTop[i]][m.genIndex[i]] = m.recoIndex[m.matchedIndex[i]];
```
kpedro88: tabs vs spaces

Utils/src/HiddenSectorProducer.cc
```diff
@@ -196,19 +237,166 @@ double HiddenSectorProducer::calculateMT2(const edm::Handle<edm::View<reco::GenM
...
+
+  //Save matched indices to final array
+  for(unsigned m = 0; m < leftovers.matchedIndex.size(); m++){
+    if (leftovers.matchedIndex[m] != -1) {
+      indices[leftovers.genIndexTop[m]][leftovers.genIndex[m]] = leftovers.recoIndex[leftovers.matchedIndex[m]];
+      tmpRecoMatched[leftovers.recoIndex[leftovers.matchedIndex[m]]] = true;
+      matchStageReco[leftovers.recoIndex[leftovers.matchedIndex[m]]] = 2;
+    }
+  }
+  recoMatched = tmpRecoMatched;
```
kpedro88: are these assignments actually needed, since we just return `indices` directly after this?

Utils/src/HiddenSectorProducer.cc
```diff
@@ -40,37 +44,50 @@ CandPtr daughter_noexcept(const T& mother, unsigned i) {
...
+    edm::InputTag JetTag_, MetTag_, GenMetTag_, GenTag_, GenJetTag_, GenIndexTag_, CandTag_;
     edm::EDGetTokenT<edm::View<pat::Jet>> JetTok_;
     edm::EDGetTokenT<edm::View<pat::MET>> MetTok_;
     edm::EDGetTokenT<edm::View<reco::GenMET>> GenMetTok_;
     edm::EDGetTokenT<edm::View<reco::GenParticle>> GenTok_;
     edm::EDGetTokenT<edm::View<reco::GenJet>> GenJetTok_;
+    edm::EDGetTokenT<edm::View<reco::Candidate>> CandTok_;
+    edm::EDGetTokenT<std::vector<int>> GenIndexTok_;
     double coneSize_;
+    edm::ParameterSet njhConfig;
```
kpedro88: member variables should end in _

Utils/src/HiddenSectorProducer.cc
```diff
@@ -367,13 +626,237 @@ void HiddenSectorProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+	}
+      }
+
+      GenJets_nConstituentsDarkHadrons->push_back(nDaus);
+
+      std::vector<CLorentzVector> tmp_darkHadrons;
+      std::vector<CLorentzVector> tmp_darkHadronJets;
+      std::vector<std::vector<CLorentzVector> > tmp_darkHadronJets_constituents;
+      std::vector<std::vector<int> > tmp_darkHadronJets_ConstituentPdgid;
+      //std::vector<std::vector<CLorentzVector> > tmp_darkHadronSubjets_constituents;
```
kpedro88: delete commented-out code

Utils/src/HiddenSectorProducer.cc
```diff
@@ -367,13 +626,237 @@ void HiddenSectorProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+      
+      int nDaus = 0;
+      std::map<CandPtr,std::vector<CandPtr>,decltype(comp)> darkHadronMap(comp);
+      for(unsigned i = 0; i < i_jet.numberOfDaughters(); ++i){
+        CandPtr dau = daughter_noexcept(i_jet,i);
+        CandPtr darkHadron = getAncestor(DarkHadronIDs_,dau);
+        if (darkHadron!=nullptr){
+          darkHadronMap[darkHadron].push_back(dau);
+	  nDaus++;
+	}
```
kpedro88: tabs vs spaces

Utils/src/HiddenSectorProducer.cc
```diff
@@ -367,13 +626,237 @@ void HiddenSectorProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+      std::vector<std::vector<int> > tmp_darkHadronJets_ConstituentPdgid;
+      //std::vector<std::vector<CLorentzVector> > tmp_darkHadronSubjets_constituents;
+      std::vector<int> tmp_darkHadronJets_multiplicity;
+      std::vector<double> tmp_darkHadronJets_tau1;
+      std::vector<double> tmp_darkHadronJets_tau2;
+      std::vector<double> tmp_darkHadronJets_tau3;
+      for(const auto& entry : darkHadronMap){
+        tmp_darkHadrons.emplace_back(entry.first->pt(),entry.first->eta(),entry.first->phi(),entry.first->energy());
+        LorentzVector tmpjet;
+	std::vector<CLorentzVector> tmpjetconstituents;
```
kpedro88: tabs vs spaces

Utils/src/HiddenSectorProducer.cc
```diff
@@ -367,13 +626,237 @@ void HiddenSectorProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+	std::vector<int> tmpjetconstituentspdgid;
+        for(const auto& dau : entry.second){
+	  tmpjet += dau->p4();
+	  tmpjetconstituents.emplace_back(dau->pt(),dau->eta(),dau->phi(),dau->energy());
+	  tmpjetconstituentspdgid.emplace_back(dau->pdgId());
+        }
+        tmp_darkHadronJets.emplace_back(tmpjet.pt(),tmpjet.eta(),tmpjet.phi(),tmpjet.energy());
+	tmp_darkHadronJets_constituents.push_back(tmpjetconstituents);
```
kpedro88: tabs vs spaces

Utils/src/HiddenSectorProducer.cc
```diff
@@ -367,13 +626,237 @@ void HiddenSectorProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+
+  //Iterate through reco jets and get matching genjets
+  int recoJetIndex = 0;
+  int nRecoJetCands = 0;
+  for(const auto& i_jet : *(h_jets.product())){
+
+    int genJetIndex = genIndex->at(recoJetIndex);
+    
+    Jets_nConstituents->push_back(jets_cands[recoJetIndex].size());
+    //Jets_constituents_pdgid->push_back(cands_pdgids[recoJetIndex]);
```
kpedro88: delete commented-out code

Utils/src/HiddenSectorProducer.cc
```diff
@@ -367,13 +626,237 @@ void HiddenSectorProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+    std::vector<std::vector<int>> matchStage;
+    std::vector<std::vector<int>> RecoPdgid;
+
+    for(unsigned i = 0; i < matchedRecoIndex.size(); i++){
+      std::vector<CLorentzVector> recoSubjetConstituents;
+      CLorentzVector tmpJet;
+      std::vector<int> tmpMatchStage;
+      std::vector<int> tmpRecoPdgid;
+      for(unsigned j = 0; j < matchedRecoIndex[i].size(); j++){
+	if (matchedRecoIndex[i][j] != -1) {
```
kpedro88: tabs vs spaces

Utils/src/HiddenSectorProducer.cc
```diff
@@ -367,13 +626,237 @@ void HiddenSectorProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+
+    //For unmatched reco particles, assign them to the closest dark hadron, and the second closest to demonstrate an uncertainty in this procedure
+    std::vector<std::vector<CLorentzVector> > recoSubjetsNextDH = recoSubjets;
+    std::vector<std::vector<CLorentzVector> > newRecoSubjets = recoSubjets;
+    std::vector<int> recoUnmatchedPdgid = {};
+    std::vector<int> isAssigned;
+    int nRecoUnmatched = 0;
+    for(unsigned i = 0; i < recoMatched.size(); i++){
+      isAssigned.push_back(0);
+      //std::cout << recoMatched[i] << std::endl;
```
kpedro88: delete commented-out code

Utils/src/HiddenSectorProducer.cc
```diff
@@ -367,13 +626,237 @@ void HiddenSectorProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+    Jets_constituents_matchStageReco->push_back(RecoMatchStage);
+
+    //For unmatched reco particles, assign them to the closest dark hadron, and the second closest to demonstrate an uncertainty in this procedure
+    std::vector<std::vector<CLorentzVector> > recoSubjetsNextDH = recoSubjets;
+    std::vector<std::vector<CLorentzVector> > newRecoSubjets = recoSubjets;
+    std::vector<int> recoUnmatchedPdgid = {};
+    std::vector<int> isAssigned;
+    int nRecoUnmatched = 0;
+    for(unsigned i = 0; i < recoMatched.size(); i++){
+      isAssigned.push_back(0);
```
kpedro88: could just do `std::vector<int> isAssigned(recoMatched.size(), 0)`? (before the loop)

Utils/src/HiddenSectorProducer.cc
```diff
@@ -367,13 +626,237 @@ void HiddenSectorProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+      //std::cout << recoMatched[i] << std::endl;
+      if (recoMatched[i] == false ) {
+    	recoUnmatchedPdgid.push_back(cands_pdgids[recoJetIndex][i]);
+	nRecoUnmatched++;
+
+	CLorentzVector recoPart = jets_cands[recoJetIndex][i];
+	std::vector<CLorentzVector> darkHadronJets = GenJets_darkHadronJets->at(genJetIndex);
```
kpedro88: `const std::vector<CLorentzVector>&` (avoid copy)

Utils/src/HiddenSectorProducer.cc
```diff
@@ -367,13 +626,237 @@ void HiddenSectorProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+	  std::iota(idx.begin(), idx.end(), 0);
+	  // sort indexes based on comparing deltaR values using std::stable_sort, now idx vector is a list of indices sorted by deltaR 
+	  std::stable_sort(idx.begin(), idx.end(), [&dhjDeltaR](int i1, int i2) {return dhjDeltaR[i1] < dhjDeltaR[i2];});	
+	  newRecoSubjets[idx[0]].push_back(jets_cands[recoJetIndex][i]);
+	  isAssigned[i] = 1;
+	  if (dhjDeltaR.size() > 1) recoSubjetsNextDH[idx[1]].push_back(jets_cands[recoJetIndex][i]);
+	}
+      }
+    }
+    double unmatchedFraction = nRecoUnmatched / jets_cands[recoJetIndex].size();
```
kpedro88: never actually used?

kpedro88:
Also, the integration test failed.

Input file: https://cernbox.cern.ch/remote.php/dav/public-files/3xHyo11H3fAgCHZ/store.mc.RunIISummer20UL16MiniAODv2.QCD_HT2000toInf_TuneCP5_PSWeights_13TeV-madgraphMLM-pythia8.MINIAODSIM.106X_mcRun2_asymptotic_v17-v1.120000.0605C02A-A789-C548-8D73-9B7D0A4D5561_100evt.root

Test command:
```
cmsRun \
  $CMSSW_BASE/src/TreeMaker/Production/test/runMakeTreeFromMiniAOD_cfg.py \
  scenario=Summer20UL16 \
  dataset=file:store.mc.RunIISummer20UL16MiniAODv2.QCD_HT2000toInf_TuneCP5_PSWeights_13TeV-madgraphMLM-pythia8.MINIAODSIM.106X_mcRun2_asymptotic_v17-v1.120000.0605C02A-A789-C548-8D73-9B7D0A4D5561_100evt.root \
  outfile=Summer20UL16.QCD_HT2000toInf_TuneCP5_PSWeights_13TeV-madgraphMLM-pythia8 \
  numevents=100
```

Error message:
```
----- Begin Fatal Exception 08-May-2025 19:26:50 UTC-----------------------
An exception of category 'StdException' occurred while
   [0] Processing  Event run: 1 lumi: 37831 event: 44488172 stream: 0
   [1] Running path 'WriteTree'
   [2] Prefetching for module TreeMaker/'TreeMaker2'
   [3] Calling method for module HiddenSectorProducer/'HiddenSector'
Exception Message:
A std::exception was thrown.
vector::_M_range_check: __n (which is 1) >= this->size() (which is 0)
----- End Fatal Exception -------------------------------------------------
```

How to debug:
```
cd $CMSSW_BASE/src/TreeMaker/Utils
touch src/HiddenSectorProducer.cc
scram b USER_CXXFLAGS="-g"
cd $CMSSW_BASE/src/TreeMaker/Production/test/
cmsTraceExceptions cmsRun ...
```

TreeMaker/interface/TreeMaker.h
easmith5: This is actually just copied from the one-level nested class, as is most of this code, but I believe that is correct

easmith5:
Ok I think I've addressed these comments! I also fixed the failed integration test by wrapping the Lund related code in the signal_ requirement. I tried to fix the spacing - the HiddenSectorProducer class seems to not be tab-everywhere, but you commented specifically on a couple locations so I changed my code to be tab based. I could modify the whole file if you prefer. Thanks!

TreeMaker/interface/TreeMaker.h
kpedro88: this one is not fixed yet

Utils/src/HiddenSectorProducer.cc
```diff
@@ -196,19 +236,162 @@ double HiddenSectorProducer::calculateMT2(const edm::Handle<edm::View<reco::GenM
...
+	// get list of gen particles for each pdgid
+	for(unsigned i = 0; i < genSubjetConstituents.size(); i++){
+		for(unsigned j = 0; j < genSubjetConstituents[i].size(); j++){
+			auto it = std::find(pdgids.begin(), pdgids.end(), genSubjetPdgid[i][j]);
+			int index = -1;
+			if ( it != pdgids.end()) {
+				index = it - pdgids.begin();
+				toMatch[index].genParts.push_back(genSubjetConstituents[i][j]);
+				toMatch[index].genIndexTop.push_back(i);
+				toMatch[index].genIndex.push_back(j);
```
kpedro88: the point is that these lines can be outside/after the if/else, because now they are exactly the same in both cases

Utils/src/HiddenSectorProducer.cc
```diff
@@ -196,19 +236,162 @@ double HiddenSectorProducer::calculateMT2(const edm::Handle<edm::View<reco::GenM
...
+	}
+
+	// get list of reco particles for each pdgid/hadrons
+	for(unsigned r = 0; r < jetCands.size(); r++){
+		auto it = std::find(pdgids.begin(), pdgids.end(), jetCandsPdgid[r]);
+		int index = -1;
+		if ( it != pdgids.end()) {
+			index = it - pdgids.begin();
+			toMatch[index].recoParts.push_back(jetCands[r]);
+			toMatch[index].recoIndex.push_back(r);
```
kpedro88: and here

kpedro88:
The spacing looks consistent enough now

TreeMaker/interface/TreeMaker.h
easmith5: Ah looks like I fixed the one in the previously existing class only. Should be done now.

Utils/src/HiddenSectorProducer.cc
easmith5: fantastic point

Utils/src/HiddenSectorProducer.cc
```diff
@@ -367,13 +618,225 @@ void HiddenSectorProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+		  }
+
+		  const auto i_genJet = h_genjets->at(genJetIndex);
+
+
+		  // Match gen to reco particles using pdgid and deltaR matching
+		  std::vector<bool> recoMatched(jets_cands[recoJetIndex].size(), -1);
+		  std::vector<int> matchStageReco(jets_cands[recoJetIndex].size(), -1);
+
+		  std::vector<std::vector<int>> matchedRecoIndex = matchParticles(GenJets_darkHadronJets_constituents->at(genJetIndex), GenJets_darkHadronJets_constituentsPdgid->at(genJetIndex), jets_cands[recoJetIndex], cands_pdgids[recoJetIndex], recoMatched, matchStageReco);
```
kpedro88: `const std::vector<std::vector<int>>&`

Utils/src/HiddenSectorProducer.cc
```diff
@@ -196,19 +236,159 @@ double HiddenSectorProducer::calculateMT2(const edm::Handle<edm::View<reco::GenM
...
+	pdgids_out.resize(h_jets->size());
+	for(unsigned c = 0; c < h_cands->size(); ++c){
+		const auto& candPtr = h_cands->ptrs()[c];
+		//if this cand is kept, it will be appended to cands_out
+		for(unsigned j = 0; j < h_jets->size(); ++j){
+			const auto& jet = h_jets->at(j);
+			const auto& daughterPtrs = jet.daughterPtrVector();
+			if(cands_out[j].size()==daughterPtrs.size()) continue;
+			bool keep = false;
+			for(const auto& candJetDau : jet.daughterPtrVector()){
```
kpedro88: use the existing `daughterPtrs` variable here instead of calling `daughterPtrVector()` again (costly)

Utils/src/HiddenSectorProducer.cc
```diff
@@ -196,19 +236,159 @@ double HiddenSectorProducer::calculateMT2(const edm::Handle<edm::View<reco::GenM
...
+					pdgids_out[j].emplace_back(candPtr->pdgId());
+					//in a given jet collection, a candidate can only be in one jet
+					break;
+				}
+			}
+			// if we are here either we broke out of candJetDau loop because we matched that jet dau, and we should skip the rest of the jets by breaking out of jet loop and going to next h_cand, or we went through all jet daughters and keep is false and we should go to next jet and see if we match there. Basically if we break out of daughters, also break out of jets loop
+			if (keep) break;
+		}
+	}
+	return;
```
kpedro88: unneeded

kpedro88:
I ran a profile on our benchmark signal sample.

Before:
* `HiddenSectorProducer` is 0.89 / 269.88 CPU ticks
    ```
    [671]       0.3       0.89       0.02 / 0.87       HiddenSectorProducer::produce(edm::StreamID, edm::Event&, edm::EventSetup const&) const
                0.2  .........       0.63 / 0.63         HiddenSectorProducer::checkLast(reco::GenJet const&, std::unordered_set<reco::Candidate const*, std::hash
                0.0  .........       0.13 / 0.13         HiddenSectorProducer::firstDark(reco::Candidate const*, std::unordered_set<reco::Candidate const*, std::h
                0.0  .........       0.04 / 0.22         HiddenSectorProducer::isParticle(std::unordered_set<unsigned int, std::hash<unsigned int>, std::equal_to<
   ```

* Size/Entry: 9.11 KiB

After:
* `HiddenSectorProducer` is 12.41 / 279.34 CPU ticks (6th largest contributor)
    ```
    [68]        3.6      12.41       0.05 / 12.36      HiddenSectorProducer::produce(edm::StreamID, edm::Event&, edm::EventSetup const&) const
                2.5  .........       8.57 / 8.57         NjettinessHelper::getTau(unsigned int, std::vector<ROOT::Math::LorentzVector<ROOT::Math::PtEtaPhiE4D<doub
                0.4  .........       1.32 / 1.32         HiddenSectorProducer::matchParticles(std::vector<std::vector<ROOT::Math::LorentzVector<ROOT::Math::PtEtaP
                0.2  .........       0.77 / 0.77         HiddenSectorProducer::matchJetsCands(edm::Handle<edm::View<pat::Jet> >&, edm::Handle<edm::View<reco::Cand
                0.2  .........       0.52 / 0.52         HiddenSectorProducer::checkLast(reco::GenJet const&, std::unordered_set<reco::Candidate const*, std::hash
                0.1  .........       0.38 / 0.38         HiddenSectorProducer::getAncestor(std::unordered_set<unsigned int, std::hash<unsigned int>, std::equal_to
                0.0  .........       0.16 / 0.16         HiddenSectorProducer::firstDark(reco::Candidate const*, std::unordered_set<reco::Candidate const*, std::h
    ```
* Size/Entry: 15.06 KiB

The main contributor is Nsubjettiness, which we can't really speed up. The biggest part of `matchParticles()` seems to be just the `matchAB()` calls, which are also needed. I found a few minor speedups here and there by looking at these parts of the code a little more closely.

The new branches ordered by size:
```
GenJetsAK8_darkHadronJets_constituents
JetsAK8_darkHadronJets_constituentsNextDH
JetsAK8_darkHadronJets_constituentsGenMatchOnly
JetsAK8_darkHadronJets_constituents
GenJetsAK8_darkHadronJets
GenJetsAK8_darkHadronJetsTau1
GenJetsAK8_darkHadronJetsTau2
GenJetsAK8_darkHadronJetsTau3
JetsAK8_darkHadronJets
GenJetsAK8_darkHadronJetsMultiplicity
```
Is there anything that could be dropped or condensed?

easmith5:
I will add these changes. Indeed as you say, I'm not sure we can do much NsubjettinessHelper. For the size of the branches, there probably is some condensing that could be done. 

```
GenJetsAK8_darkHadronJets_constituents - gen particles clustered by dark hadron ancestors
JetsAK8_darkHadronJets_constituentsNextDH - matched particles + assigned particles to the closest dark hadron 
JetsAK8_darkHadronJets_constituentsGenMatchOnly - matched particles only
JetsAK8_darkHadronJets_constituents  - matched particles + assigned particles
GenJetsAK8_darkHadronJets - dark hadron jets computed from all daughters 
GenJetsAK8_darkHadronJetsTau1 
GenJetsAK8_darkHadronJetsTau2
GenJetsAK8_darkHadronJetsTau3
JetsAK8_darkHadronJets - dark hadron jets computer from all candidates matched to gen dark hadron daughters
GenJetsAK8_darkHadronJetsMultiplicity - number of constituents which are daughters of each dark hadron
``` 

It is very much the case that `JetsAK8_darkHadronJets_constituentsNextDH`, `JetsAK8_darkHadronJets_constituentsGenMatchOnly`, `JetsAK8_darkHadronJets_constituents` could be condensed, as the GenMatchOnly is essentially a subset of the other two. I could instead of storing the full lists again, store only the additional particles, and then combine them in coffea afterwards. I think it is useful to keep all of the different configurations available to be accessed separately as it may be useful for debugging, or validation of the process. The `JetsAK8_darkHadronJets`/`GenJetsAK8_darkHadronJets` is required by coffea to be able to access the `*_darkHadronJets_constituents*` branches properly with the coffea nesting. 

Ultimately non of the Tau branches are required for the actual reweighting process but I think they will be necessary for the documentation of the procedure. 

`GenJetsAK8_darkHadronJetsMultiplicity` could probably be dropped, one can just look at the length along the innermost axis of the constituent branches to get the number of constituents in the darkHadronJets. 

I suppose there is a workflow in which we store only the indices of the constituent particles and this would reduce size greatly (similar to other producers), but introduce some complexity into the following steps. Let me know how you think I should proceed - a 40% increase is fairly significant.

kpedro88:
Here are some possibilities to reduce the disk usage:
1. Just store indices into the existing JetsConstituents and GenParticles branches. This would mean HiddenSectorProducer would consume some output from JetsConstituents and GenParticlesProducer. Probably, the best approach would be for those producers to output transient vectors (i.e. not stored in the final output file) that map the index in the miniAOD collection to the index in the TreeMaker branch. (Then HiddenSectorProducer can just track the indices during its loops.)
2. If the above seems like it may take too long to implement right away, some duplication could still be reduced by storing just one set of `JetsAK8_darkHadronJets_constituents` along with some boolean masks or integers to keep track of the different properties (match stage, which dark hadron jet they're assigned to). One could even combine this option with the first option to further reduce duplication.

These would both require some changes in how everything is reassembled in coffea. But that might actually improve the analysis processing (just loading one set of constituents in memory, instead of several copies of the same constituents).

One more possibility for Nsubjettiness: it is now actually exposed in the Python version of fastjet, as of the latest release https://github.com/scikit-hep/fastjet/releases/tag/v3.4.3.1. So we could eschew that entirely in TreeMaker and just run it as a spot check on the final output...

easmith5:
For 1. I think this is certainly possible, but interacting with the output of other producers and validating the procedure would definitely slow things down, as I've already spent some time validating the existing setup. 

2 is definitely doable fairly quickly. I've basically implemented it already and I just want to do some checks on the output. I will also check that Nsubjettiness works and looks as expected in the python version of fastjet.  

2 is fairly easy to adjust for on the coffea side, it just requires implementing a couple of masks.  1. would potentially be a bit more difficult though ultimately it is also probably just a matter of reindexing/masking, though perhaps a more delicate one.

easmith5:
Okay. I have reduced the output size significantly. I didn't run a full profile, but just comparing output file sizes, this has helped dramatically. Removing nsubjettinesshelper should also decrease the time as well. Let me know if you see any additional issues!

TreeMaker/interface/TreeMaker.h
```diff
@@ -222,7 +223,7 @@ typedef TreeObject<vector<vector<math::PtEtaPhiELorentzVector>>,vector<vector<ma
 typedef TreeObject<vector<vector<vector<math::PtEtaPhiELorentzVector>>>,vector<vector<vector<math::PtEtaPhiELorentzVectorF>>>> TreeObjectVVVLVToF;
 typedef TreeObject<vector<vector<math::XYZVector>>,vector<vector<math::XYZVectorF>>> TreeObjectVVXYZVToF;
 typedef TreeObject<vector<vector<math::XYZPoint>>,vector<vector<math::XYZPointF>>> TreeObjectVVXYZPToF;
-//typedefs for doubles
+//typedefs for doublesx
```
kpedro88: fix this typo again

Utils/src/HiddenSectorProducer.cc
```diff
@@ -80,13 +79,10 @@ class HiddenSectorProducer : public edm::global::EDProducer<edm::StreamCache<Nje
     double coneSize_;
     edm::ParameterSet njhConfig_;
     PidSet DarkSMediatorIDs_, DarkTMediatorIDs_, DarkQuarkIDs_, DarkHadronIDs_, DarkGluonIDs_, DarkStableIDs_, DarkFirstIDs_, SMQuarkIDs_;
-    std::unique_ptr<NjettinessHelper> beginStream(edm::StreamID) const {
```
kpedro88: to be sure, the usage of the Python FastJet Nsubjettiness has been validated now?

Utils/src/HiddenSectorProducer.cc
```diff
@@ -259,11 +254,8 @@ void HiddenSectorProducer::matchJetsCands(edm::Handle<edm::View<pat::Jet>>& h_je
 					break;
 				}
 			}
-			// if we are here either we broke out of candJetDau loop because we matched that jet dau, and we should skip the rest of the jets by breaking out of jet loop and going to next h_cand, or we went through all jet daughters and keep is false and we should go to next jet and see if we match there. Basically if we break out of daughters, also break out of jets loop
```
kpedro88: why was this removed?

Utils/src/HiddenSectorProducer.cc
```diff
@@ -326,14 +317,16 @@ std::vector<std::vector<int> > HiddenSectorProducer::matchParticles(std::vector<
 
 	}
 
-	for( auto& m: toMatch ) {
+	for( unsigned j = 0; j < toMatch.size(); j++ ) {
+		matchInfo m = toMatch[j];
```
kpedro88, line 330: `matchInfo& m` to avoid copying

Utils/src/HiddenSectorProducer.cc
```diff
@@ -792,17 +765,24 @@ void HiddenSectorProducer::produce(edm::StreamID iID, edm::Event& iEvent, const
...
 					  // sort indexes based on comparing deltaR values using std::stable_sort, now idx vector is a list of indices sorted by deltaR
 					  std::stable_sort(idx.begin(), idx.end(), [&dhjDeltaR](int i1, int i2) {return dhjDeltaR[i1] < dhjDeltaR[i2];});
-					  newRecoSubjets[idx[0]].push_back(jets_cands[recoJetIndex][i]);
-					  isAssigned[i] = 1;
-					  if (dhjDeltaR.size() > 1) recoSubjetsNextDH[idx[1]].push_back(jets_cands[recoJetIndex][i]);
+					  Jets_darkHadronJets_constituents->at(recoJetIndex)[idx[0]].push_back(jets_cands[recoJetIndex][i]);
+					  Jets_darkHadronJets_constituentsAssignedFirst->at(recoJetIndex)[idx[0]].push_back(1);
+					  Jets_darkHadronJets_constituentsAssignedSecond->at(recoJetIndex)[idx[0]].push_back(0);
+					  Jets_darkHadronJets_constituentsPdgid->at(recoJetIndex)[idx[0]].push_back(cands_pdgids[recoJetIndex][i]);
+					  Jets_darkHadronJets_constituentsMatchStage->at(recoJetIndex)[idx[0]].push_back(3);
```
kpedro88: is there a legend describing the match stage values somewhere? (preferably, these should be written as named constants rather than magic numbers)

Utils/src/HiddenSectorProducer.cc
easmith5: Ah I thought this was request - I guess you just meant the return statement?

Utils/src/HiddenSectorProducer.cc
easmith5: So it hasn't yet been fully validated due to the coffea issues. I will do an initial check as best I can soon.

Utils/src/HiddenSectorProducer.cc
easmith5: I haven't added a legend somewhere. When you say named constants you just mean something like: 
```
const int egMatchStage = 0
const int hadronMatchStage = 1
```
etc?

Utils/src/HiddenSectorProducer.cc
kpedro88: Yes, just the very last return statement was unnecessary. I think skipping the loop breaking is incorrect.

Utils/src/HiddenSectorProducer.cc
kpedro88: yes (or an enum, either way)

easmith5: Is putting this with the typedefs at the top of the file appropriate?

kpedro88: Best to make them members of the class to avoid the (slight) possibility of global contamination

kpedro88:
New profile:
* `HiddenSectorProducer` is 3.69 / 269.22 CPU ticks
    ```
    [178]       1.1       3.69       0.08 / 3.60       HiddenSectorProducer::produce(edm::StreamID, edm::Event&, edm::EventSetup const&) const
                0.4  .........       1.33 / 1.33         HiddenSectorProducer::matchParticles(std::vector<std::vector<ROOT::Math::LorentzVector<ROOT::Math::PtEtaP
                0.2  .........       0.68 / 0.68         HiddenSectorProducer::matchJetsCands(edm::Handle<edm::View<pat::Jet> >&, edm::Handle<edm::View<reco::Cand
                0.2  .........       0.52 / 0.52         HiddenSectorProducer::checkLast(reco::GenJet const&, std::unordered_set<reco::Candidate const*, std::hash
                0.1  .........       0.32 / 0.32         HiddenSectorProducer::getAncestor(std::unordered_set<unsigned int, std::hash<unsigned int>, std::equal_to
                0.0  .........       0.12 / 0.12         HiddenSectorProducer::firstDark(reco::Candidate const*, std::unordered_set<reco::Candidate const*, std::h
    ```
* Size/Entry: 13.71 KiB

So there is some improvement in the size, but still a lot of growth vs. the baseline. It should be acceptable for now, but for Run 3 usage, the more parsimonious approach of just storing indices into a central PFCandidates collection should be implemented.

Utils/src/HiddenSectorProducer.cc
```diff
@@ -273,11 +477,48 @@ void HiddenSectorProducer::produce(edm::StreamID, edm::Event& iEvent, const edm:
...
+
+  auto GenJets_darkHadronJets_multiplicity = std::make_unique<std::vector<std::vector<int>>>();
+  auto Jets_nConstituents_unmatched_pdgid = std::make_unique<std::vector<std::vector<int>>>();
+
+  auto Jets_constituents_pdgid = std::make_unique<std::vector<std::vector<int>>>();
+  auto GenJets_constituents_pdgid = std::make_unique<std::vector<std::vector<int>>>();
+
+  auto GenJets_darkHadronJets_tau1 = std::make_unique<std::vector<std::vector<double>>>();
+  auto GenJets_darkHadronJets_tau2 = std::make_unique<std::vector<std::vector<double>>>();
+  auto GenJets_darkHadronJets_tau3 = std::make_unique<std::vector<std::vector<double>>>();
```
kpedro88: these lines can be removed since Nsubjettiness is no longer used

kpedro88:
For the Run 3 rewrite, one possible plan (to reuse existing functionality):
1. output a collection of dark hadron jets (actual jet objects) and value maps of additional per-constituent properties (match stage etc.)
2. include that collection in the list of inputs to `JetsConstituents` (which already makes the lists of constituents by index for the regular jet collections) and allow access to the value maps
3. output a per-jet list of dark hadron jet indices (rather than the current nested vector of dark hadron jet 4-vectors)

so the final stored output would look like:
DarkHadronJets
DarkHadronJets_[property]
DarkHadronJets_constituentsIndex
JetsAK8_darkHadronJetsIndex
JetsConstituents_[property]
+ summary diagnostic info (NConstituentsUnmatched etc.)

Then the same thing should be done for GenParticles, but that will take a bit more work, since the GenParticle selection is more complicated... some `JetsConstituents`-like algorithm will have to be added to `GenParticlesProducer`. (Alternatively, `GenParticlesProducer` could be modified just to produce a mask for the entire collection, which could then be consumed by `GenJetsConstituents` as (mostly) a clone of `JetsConstituents`.)

easmith5:
This sounds good to me. Checking the size from my end, I get the same value of ~13.7 kB / entry for the current implementation. I would like to note that the original DarkJets branch started with ~11.6 kB / entry. I agree that in the long term (for Run 3 iteration) it would be best to use an index based system, though my understanding was that much of the code would be restructured for Run 3 anyways. 

I am currently working on the check of fastjet nsubjettiness - this requires awkward >=2 which requires coffea >=2023. I am in the process of updating the newer coffea treemaker schema to be able to run fastjet nsubjettiness on the darkHadronJet constituents.  

Beyond this, everything looks good to me. I can run the full diagnostic and Lund Reweighting python pipeline on these outputs and get values for the per-jet weights.

